PR #2, opened 2020-02-26
Fix Installer to create ark alias

### LINKED ISSUE #1: get.sh not creating Alias
The alias should be created by get.sh

There is a naming issue, it was called ALIAS but the script was looking for ALIAS_NAME

PR Incoming

Waterdrips:
The ALIAS var in get.sh was not being used. It was looking for
ALIAS_NAME, so I have updated to that.

This was tested with `sudo sh get.sh`.

Alias was created correctly

Fixes #1 

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -6,7 +6,7 @@
 #########################
 
 export VERIFY_CHECKSUM=0
-export ALIAS="ark"
+export ALIAS_NAME="ark"
 export OWNER=alexellis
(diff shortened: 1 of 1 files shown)


alexellis:
I wonder if this is broken elsewhere too? faas-cli etc?

Waterdrips:
faas-cli looks fine. it used ALIAS_NAME in the definition

Issue #25, opened 2020-03-01
Istio installs unsupported version


howardjohn:
## Expected Behaviour
Applications should install supported versions

## Current Behaviour

https://github.com/alexellis/arkade/blob/master/cmd/apps/istio_app.go#L77 and unsupported version of Istio is installed: https://istio.io/news/support/announcing-1.3-eol-final/
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* What Kubernetes distribution are you using?

```
kubectl version
```

* Operating System and version (e.g. Linux, Windows, MacOS):

```
uname -a

cat /etc/os-release
```

* What arkade version is this?

```
arkade version
```

aidun:
I have to install a x86 cluster to test. I think I can make the required tests this afternoon or someone else would check my PR

alexellis:
Thanks for your interest @howardjohn 👍

howardjohn:
Thanks for updating!

PR #26, opened 2020-03-02
bumped istio to 1.4.5 according to the docs to fix issue #25

aidun:
bumped istio to 1.4.5 according to the docs to fix issue #25

## Description
bumped istio to 1.4.5 according to the docs to fix issue #25

## Motivation and Context

## How Has This Been Tested?

kind create cluster
arkade install istio
helm list -A
kubectl get all -n Istio-system

--> processes are running
--> helm charts in 1.4.5 are present

## Types of changes

## Checklist:

### DIFF
--- a/cmd/apps/istio_app.go
+++ b/cmd/apps/istio_app.go
@@ -74,7 +74,7 @@ func MakeInstallIstio() *cobra.Command {
 			return err
 		}
 
-		istioVer := "1.3.3"
+		istioVer := "1.4.5"
 
 		err = addHelmRepo("istio", "https://storage.googleapis.com/istio-release/releases/"+istioVer+"/charts", helm3)
 		if err != nil {
@@ -114,11 +114,11 @@ func MakeInstallIstio() *cobra.Command {
 		outputPath := path.Join(chartPath, "istio")
 
 		if initIstio, _ := command.Flags().GetBool("init"); initIstio {
-			err = helm3Upgrade(outputPath, "istio/istio-init", namespace, "", defaultVersion, overrides, wait)
+			// Waiting for the crds to appear
+			err = helm3Upgrade(outputPath, "istio/istio-init", namespace, "", defaultVersion, overrides, true)
 			if err != nil {
 				return fmt.Errorf("unable to istio-init install chart with helm %s", err)
 			}
-
 		}
 
 		customFlags, customFlagErr := command.Flags().GetStringArray("set")


aidun:
Need x86 test k8s to test (only have rapis currently). Maybe someone can confirm that it worked.

aidun:
Testet it with Kind. Is working as expected.

cmd/apps/istio_app.go
```diff
@@ -114,11 +114,21 @@ func MakeInstallIstio() *cobra.Command {
...
 
 		if initIstio, _ := command.Flags().GetBool("init"); initIstio {
-			err = helm3Upgrade(outputPath, "istio/istio-init", namespace, "", defaultVersion, overrides, wait)
+			// Waiting for the crds to appear
+			err = helm3Upgrade(outputPath, "istio/istio-init", namespace, "", defaultVersion, overrides, true)
 			if err != nil {
 				return fmt.Errorf("unable to istio-init install chart with helm %s", err)
 			}
 
+			err = kubectl("-n", "istio-system", "wait", "--for=condition=complete", "job", "--all")
```
alexellis: Why is this required? Isn't it part of the `--wait` flag that is optional on apps like `openfaas` and `inlets-operator`?

cmd/apps/istio_app.go
```diff
@@ -114,11 +114,21 @@ func MakeInstallIstio() *cobra.Command {
...
+
+			if err != nil {
+				return fmt.Errorf("error waiting for the crds: %s", err)
```
alexellis: Should be: `CRDs`

cmd/apps/istio_app.go
```diff
@@ -114,11 +114,21 @@ func MakeInstallIstio() *cobra.Command {
...
+			}
+		}
+
+		if err != nil {
```
alexellis: Why do we have the same error twice, is this a different step?

cmd/apps/istio_app.go
```diff
@@ -114,11 +114,21 @@ func MakeInstallIstio() *cobra.Command {
 		outputPath := path.Join(chartPath, "istio")
 
 		if initIstio, _ := command.Flags().GetBool("init"); initIstio {
-			err = helm3Upgrade(outputPath, "istio/istio-init", namespace, "", defaultVersion, overrides, wait)
+			// Waiting for the crds to appear
+			err = helm3Upgrade(outputPath, "istio/istio-init", namespace, "", defaultVersion, overrides, true)
```
alexellis: Please use the wait flag instead of forcing wait to be true.

I'm OK if you want to add a flag and set it to default a true value. See the other apps for examples or ask in #arkade on Slack if you need help with that.

cmd/apps/istio_app.go
aidun: This are two different points.

I will remove the `kubectl` command, it is not needed, but we have to force wait here, if not the second install step will fail. Is it okay for you?

cmd/apps/istio_app.go
aidun: Done

cmd/apps/istio_app.go
aidun: If I don´t force and to wait, the second step will always fail. I think its a bad behavior.

cmd/apps/istio_app.go
aidun: i removed it. copy paste error.

aidun:
Thank you for the feedback. I followed your comments and fixed the most of it. I have only concerns to implement the optional wait on the first step. I added my explanation to your inline review.

PR #27, opened 2020-03-02
WIP: Added helm3 option to chart_app

### LINKED ISSUE #20: Add --helm3 flag to chart app
## Description

Add --helm3 flag to chart app

The `arkade install chart` command should have helm3 support and use it by default.

kadern0:
Signed-off-by: kadern0 <kaderno@gmail.com>

Fixes #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Status](https://travis-ci.com/alexellis/arkade.svg?branch=master)](https://travi
 [![GoDoc](https://godoc.org/github.com/alexellis/arkade?status.svg)](https://godoc.org/github.com/alexellis/arkade)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 [![Go Report Card](https://goreportcard.com/badge/github.com/alexellis/arkade)](https://goreportcard.com/report/github.com/alexellis/arkade)
-<!-- ![GitHub All Releases](https://img.shields.io/github/downloads/alexellis/arkade/total) -->
+![GitHub All Releases](https://img.shields.io/github/downloads/alexellis/arkade/total)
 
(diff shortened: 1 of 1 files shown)


aidun:
The code looks okay, without running it. I will check it tomorrow. I think you have a problem with go.mod there is a dependency to k8s which is not needed, you should remove it. Maybe you have also an issue with go fmt, Travis complains about something.

cmd/apps/chart_app.go
```diff
@@ -131,12 +136,25 @@ before using the generic helm chart installer command.`,
...
+			}
 
+		} else {
+			outputPath = path.Join(chartPath, "chart/rendered")
+			err = templateChart(chartPath, chartName, namespace, outputPath, "values.yaml", setMap)
+			if err != nil {
+				return err
+			}
+		}
 		err = kubectl("apply", "--namespace", namespace, "-R", "-f", outputPath)
```
aidun: This part should be only for helm2. The helm 3 upgrade don't need it and is causing the strange error.

go.mod
```diff
@@ -8,4 +8,5 @@ require (
 	github.com/sethvargo/go-password v0.1.2
 	github.com/spf13/cobra v0.0.5
 	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
+	k8s.io/apimachinery v0.17.3
```
aidun: This dependency should be removed. `go mod tidy` can fix it.

cmd/apps/chart_app.go
aidun: the `kubectl apply -f` is meant here.

go.mod
kadern0: Thank you for your help, it seems to be working fine now. I've submitted a new PR. 

I cannot @ you for some reason.

PR #30, opened 2020-03-03
Fixes #6: Added grafana app

### LINKED ISSUE #6: [App request] Grafana
## App

Grafana - for dashboards and monitoring

Name `grafana`

Helm: 3 only

Persistence off by default, but flag available

https://github.com/helm/charts/tree/master/stable/grafana

aidun:
This PR Fixes #6 

- helm3 only
- default without persistence
- persistence could be enabled with --persistence

Signed-off-by: Markus Hartmann <markush1986@gmail.com>

## Description

Installs Grafana via helm3.

## Motivation and Context

## How Has This Been Tested?

```
kind create cluster
arkade install gafana
# Show all resources
kubectl get all -n grafana

arkade install gafana --persistence
# show the bound pvc
kubectl get pvc -n Grafana
```
## Types of changes

## Checklist:

### DIFF
--- a/cmd/apps/grafana_app.go
+++ b/cmd/apps/grafana_app.go
@@ -1,0 +1,134 @@
+package apps
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path"
+
+	"github.com/alexellis/arkade/pkg"
+	"github.com/alexellis/arkade/pkg/config"
+	"github.com/alexellis/arkade/pkg/env"
+	"github.com/alexellis/arkade/pkg/helm"
+	"github.com/spf13/cobra"
+)
+
+func MakeInstallGrafana() *cobra.Command {
+	var grafana = &cobra.Command{
+		Use:          "grafana",
+		Short:        "Install grafana",
+		Long:         "Install grafana for creating dashboards",
+		Example:      "arkade install grafana",
+		SilenceUsage: true,
+	}
+
+	grafana.Flags().StringP("namespace", "n", "grafana", "The namespace to install grafana")
+	grafana.Flags().Bool("update-repo", true, "Update the helm repo")
+	grafana.Flags().Bool("persistence", false, "Make grafana persistent")
+
+	grafana.RunE = func(command *cobra.Command, args []string) error {
+
+		const chartVersion = "5.0.4"
+
+		// Get all flags
+		wait, _ := command.Flags().GetBool("wait")
+		persistence, _ := command.Flags().GetBool("persistence")
+		namespace, _ := command.Flags().GetString("namespace")
+
+		// initialize client env
+		userPath, err := config.InitUserDir()
+		if err != nil {
+			return err
+		}
+
+		clientArch, clientOS := env.GetClientArch()
+
+		log.Printf("Client: %s, %s\n", clientArch, clientOS)
+		log.Printf("User dir established as: %s\n", userPath)
+
+		os.Setenv("HELM_HOME", path.Join(userPath, ".helm"))
+
+		_, err = helm.TryDownloadHelm(userPath, clientArch, clientOS, true)
+		if err != nil {
+			return err
+		}
+
+		// Update chart repositories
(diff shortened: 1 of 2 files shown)


alexellis:
The PVC should be off by default, thanks for handling that 👍 

Is there ARM support?

aidun:
If you disable sidecar and dashboard support:
```
      sidecar:
        datasources:
          enabled: false
        dasboards:
          enabled: false
```
This images have no arm support.

alexellis:
In which case, don't you think we should prevent the installation on an ARM cluster? See the Istio app as an example.

aidun:
I don't think so, we should disable the features, because we don't use them right now.

alexellis:
Please can you confirm the following is true: this app will not install to ARM? If so, then can you set up platform filtering as per the Istio app?

alexellis:
@Waterdrips Cc

Waterdrips:
This will be very useful! I wonder if we need a `grafana-dashboard` app that can install a grafana dashboard by it's id

aidun:
So if you want this feature , I will exit with errors on non amd64 platforms.  I think we can add the dashboards via flags `—dashboardid`, `—dashboardurl` and `—dashboardfile`.

What about datasources?

aidun:
> Please can you confirm the following is true: this app will not install to ARM? If so, then can you set up platform filtering as per the Istio app?

I can confirm this. The grafana pod do not start because of this sidecar containers. If you disable them, grafana is arm ready.

alexellis:
I don't know how people would want to use this app, I would assume simply installing without sidecars on any platform would be useful, especially for RPi users (from what I am reading, it sounds like that works?)

aidun:
My suggestion: I will implement the original use case. In an additional step, I will do a second PR to provide dashboard and datasource functionality. This will take some time, because I don't know much about Grafana automation. In this time I try to find the maintainer of the sidecar image and ask them to provide arm images. If I find images good, if not this features only work on amd64.

aidun:
I pushed the PR with the original scope.

alexellis:
Hi @aidun due to merging another PR we need a rebase. Please could you look at that? Thanks

aidun:
@alexellis is rebased

cmd/apps/grafana_app.go
```diff
@@ -0,0 +1,134 @@
...
+		fmt.Println(grafanaInstallMsg)
+
+		return nil
+	}
+
+	return grafana
+}
+
+const grafanaInstallMsg = `=======================================================================
+= grafana has been installed.                                   =
```
alexellis: I don't think the ASCII lines up properly here?

cmd/apps/grafana_app.go
```diff
@@ -0,0 +1,134 @@
...
+=======================================================================
+
+Get the admin password:
+
+	kubectl get secret --namespace grafana grafana -o jsonpath="{.data.admin-password}" | base64 --decode ; echo
```
alexellis: Please only indent `  ` 2x spaces.

cmd/apps/grafana_app.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return grafana
+}
+
+const grafanaInstallMsg = `=======================================================================
+= grafana has been installed.                                   =
+=======================================================================
+
+Get the admin password:
```
alexellis: These should be prefixed with `#` as per the other apps.

cmd/apps/grafana_app.go
alexellis: centre the text please

alexellis:
Thanks for that, approved for merge. You just need to rebase it now - conflict is on cmd/install.go

```
git pull origin master --rebase
```

Note that `--rebase` is really important here, we can't merge merge commits.

alexellis:
@aidun quick question, where is the info message?

aidun:
I missed it. I will add it.

Issue #88, opened 2020-04-08
ark alias conflicts with Ark, the KDE archiving tool


PierreRustOrange:
## Expected Behaviour
When  arkade is installed, my os basic tools should keep working... ;)

## Current Behaviour
When arkade is installed, it creates a symlink `/usr/local/bin/ark` that override the `ark` utility ( located in `/usr/bin/ark`) which is used to handle all archive file on KDE desktops (tar, bz, etc.). All gui-based operations on these files are then broken (from the file manager, etc.).

## Possible Solution
Avoid creating that alias by default, use a command line argument to create it only when explicitly requested.

alexellis:
Hi @PierreRustOrange would it be useful to you if we have an env-var to suppress the alias?

```
export CREATE_ARKADE_ALIAS=0
```

PierreRustOrange:
Thanks for your answer.

I'd rather have an env-var to create the alias than to suppress it,  it's actually very easy to fix the problem (only need to rm the symlink), the only difficult part is to identify it. It took me some time to understand why I could not open any tar.gz anymore...
Either way, as long as it is mentioned in the doc I'd say it's fine !

alexellis:
What if we checked for ark existing in PATH before creating it? Would that resolve this issue?

PierreRustOrange:
That's a great idea, that would completely resolve the issue and could probably be implemented by simply checking the return form `whereis ark` before creating the symlink.

alexellis:
I know you didn't come here necessary to write a PR, but would you be up for giving it a shot? we use get.sh in this repo.

PierreRustOrange:
You're right, I should have done that before.
The PR is done, my bash is probably a bit rusty, there might be a more efficient way to implement that check but it seems to do the trick.

Waterdrips:
/close
This was implemented in #89 
thanks

PR #89, opened 2020-04-09
Avoid creating a conflicting alias

PierreRustOrange:
This fixes issue #88 .

Please note that while I tested it locally, I did not perform an end-to-end test, as the get.sh script downloads directly from github.

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -162,9 +162,8 @@ getPackage() {
                 rm "$targetFile"
             fi
 
-            existingAlias=$(whereis -b $ALIAS_NAME | cut -f 2 -d :)
-            if [ ! -z "${existingAlias// }" ]; then
-                echo "There is already a command '${existingAlias// }' in the path, do NOT create alias"
+            if [ $(which $ALIAS_NAME) ]; then
+                echo "There is already a command '$ALIAS_NAME' in the path, do NOT create alias"
             else
                 if [ -n "$ALIAS_NAME" ]; then
                     if [ ! -L $BINLOCATION/$ALIAS_NAME ]; then
--- a/get.sh
+++ b/get.sh
@@ -163,7 +163,7 @@ getPackage() {
             fi
 
             if [ $(which $ALIAS_NAME) ]; then
-                echo "There is already a command '$ALIAS_NAME' in the path, do NOT create alias"
+                echo "There is already a command '$ALIAS_NAME' in the path, NOT creating alias"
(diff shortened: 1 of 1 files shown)


alexellis:
Thanks for the PR. Just need to check that `whereis` exists on Git Bash, MacOS and Linux (bash)

alexellis:
Thanks to @mehyedes, it seems that `whereis` is not supported in Git Bash, so perhaps `where` could substitute? https://stackoverflow.com/questions/54485435/git-bash-whereis-command-not-found

mehyedes:
Using `which` would be better here IMHO, as it is available also on Git Bash.
Besides, we're just trying to find out if the alias exists, so the format of the output is not that important for us.
We can use the exit code of the `which` command, to see if it ran successfully by checking the value of `$?`. If it returns `1`, it means the symlink does not exist in the path:
```
$ ln -nfs /usr/bin/docker /usr/bin/docker-alias
$ which docker-alias
/bin/docker-alias
$ echo $?
0
```
That should also work on Git Bash

PierreRustOrange:
OK, you're right, `which` is enough and it makes the code simpler. I've pushed the corresponding modification.

get.sh
```diff
@@ -162,10 +162,14 @@ getPackage() {
                 rm "$targetFile"
             fi
 
-            if [ -n "$ALIAS_NAME" ]; then
-                if [ ! -L $BINLOCATION/$ALIAS_NAME ]; then
-                    ln -s $BINLOCATION/$REPO $BINLOCATION/$ALIAS_NAME
-                    echo "Creating alias '$ALIAS_NAME' for '$REPO'."
+            if [ $(which $ALIAS_NAME) ]; then
```
mehyedes, line 165: 👍

get.sh
```diff
@@ -162,10 +162,14 @@ getPackage() {
...
+                echo "There is already a command '$ALIAS_NAME' in the path, do NOT create alias"
```
mehyedes, line 166: Could you also change the message, just so that it's consistent with the other one ?
```suggestion
                echo "There is already a command '$ALIAS_NAME' in the path, NOT creating alias"
```

mehyedes:
Thanks ! Looks fine now 👍

Waterdrips:
LGTM. Tested on Linux.

alexellis (approved):
Approved

PR #97, opened 2020-04-13
Download (get) commands and CLIs

alexellis:
Initial templating for kubectl and faas-cli patterns to
validate if the rules can be encoded into Go templates for
download URL patterns.

Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

### DIFF
--- a/pkg/get/get.go
+++ b/pkg/get/get.go
@@ -1,0 +1,229 @@
+package get
+
+import (
+	"bytes"
+	"crypto/tls"
+	"fmt"
+	"html/template"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+)
+
+type Tool struct {
+	Name           string
+	Repo           string
+	Owner          string
+	Version        string
+	URLTemplate    string
+	BinaryTemplate string
+}
+
+var templateFuncs = map[string]interface{}{
+	"HasPrefix": func(s, prefix string) bool { return strings.HasPrefix(s, prefix) },
+}
+
+// Download fetches the download URL for a release of a tool
+// for a given os,  architecture and version
+func GetDownloadURL(tool *Tool, os, arch, version string) (string, error) {
+	ver := tool.Version
+	if len(version) > 0 {
+		ver = version
+	}
+
+	dlURL, err := tool.GetURL(os, arch, ver)
+	if err != nil {
+		return "", err
+	}
+
+	return dlURL, nil
+}
+
+func (tool Tool) GetURL(os, arch, version string) (string, error) {
+
+	if len(tool.URLTemplate) == 0 {
+		return getURLByGithubTemplate(tool, os, arch, version)
+	}
+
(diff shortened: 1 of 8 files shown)


alexellis:
@Waterdrips PTAL?

pkg/get/get.go
```diff
@@ -0,0 +1,194 @@
...
+	return version, nil
+}
+
+func MakeTools() []Tool {
+	tools := []Tool{
+		{
+			Owner: "openfaas",
+			Repo:  "faas-cli",
+			Name:  "faas-cli",
+			BinaryTemplate: `{{ if HasPrefix .OS "ming" -}}
```
alexellis, line 136: This works fairly well with GitHub

pkg/get/get.go
```diff
@@ -0,0 +1,194 @@
...
+{{- else if eq .Arch "aarch64" -}}
+{{.Name}}-arm64
+{{- end -}}`,
+		},
+		{
+			Owner:       "kubernetes",
+			Repo:        "kubernetes",
+			Name:        "kubectl",
+			Version:     "v1.18.0",
+			URLTemplate: `https://storage.googleapis.com/kubernetes-release/release/{{.Version}}/bin/{{.OS}}/{{.Arch}}/kubectl`,
```
alexellis, line 153: This doesn't work so well with building arbitrary URLs - I need to run a function to determine the Arch value i.e. `amd64` for 64-bit MacOS. I am now thinking about having a `func` instead.

Originally the idea of a template was to be able to have this structure on disk and not compiled-in, using a `func` would prevent that.

pkg/get/get.go
```diff
@@ -0,0 +1,194 @@
...
+	if res.Body != nil {
+		defer res.Body.Close()
+	}
+	if res.StatusCode != 302 {
+		return "", fmt.Errorf("incorrect status code: %d", res.StatusCode)
+	}
+
+	loc := res.Header.Get("Location")
+	if len(loc) == 0 {
+		return "", fmt.Errorf("unable to determine release of kubeseal")
```
Waterdrips, line 124: kubeseal --> "tool"

pkg/get/get.go
alexellis: Seems like you can have a more complex template i.e.

```
`{{$arch := "arm"}}
{{- if eq .Arch "x86_64" -}}
{{$arch = "amd64"}}
{{- end -}}
			https://storage.googleapis.com/kubernetes-release/release/{{.Version}}/bin/{{.OS}}/{{$arch}}/kubectl`
```

cc @LucasRoesler @Waterdrips

LucasRoesler: if you want it on disk, then i think using an gotmpl input is required. The simplest thing is to probably just pass the Tool object in as the template context

PR #130, opened 2020-06-19
Support arkade get with tgz/zip archives

### LINKED ISSUE #123: [Tools] - arkade get should be able to cope with tgz/zip and multi-file archives
Today arkade can download a single binary from a release page using a Go template to form the URL.

```
# arkade get faas-cli

Downloading faas-cli
https://github.com/openfaas/faas-cli/releases/download/0.12.4/faas-cli-darwin
Tool written to: /var/folders/p7/ptxtv_pd3n12fd6mc5wjhk1h0000gn/T/faas-cli-darwin

Run the following to copy to install the tool:

chmod +x /var/folders/p7/ptxtv_pd3n12fd6mc5wjhk1h0000gn/T/faas-cli-darwin
sudo install -m 755 /var/folders/p7/ptxtv_pd3n12fd6mc5wjhk1h0000gn/T/faas-cli-darwin /usr/local/bin/faas-cli
```

Helm, however ships as a compressed multi-file archive with the binary inside. We do download this automatically via `arkade install` but the behaviour should be adapted for `arkade get` so that we can add helm and similar CLIs i.e. `arkade get helm` downloads the helm binary.

kubectx now ships as a single-file archive with a single binary inside, this is a simpler case, but still involves more code paths and conditional statements. Until this feature is completed, we can just use the bash variant which is going to exist for a while yet and is not compressed.

Ideally we should support:

* single binary - (faas-cli, kubectx)
* single binary - compressed (tar.gz) (kubectx)
* multi-file archive - where we want a single binary to be pulled out of it (helm)

Changes will be required in:

https://github.com/alexellis/arkade/blob/master/cmd/get.go

And

https://github.com/alexellis/arkade/blob/master/pkg/get/get.go

tuananh:
Signed-off-by: Tuan Anh Tran <me@tuananh.org>

Allow `arkade get` to download tgz/zip archives

There are a few points I still need clarification.

* for kubectx, the cli name is same as the tool so I can reuse `tool.Name`. However, will there be case the cli name is different than the repo name? In that case, should I add a new field `BinaryName` to the Tool struct and use that to pull the cli from the tgz/zip archive instead? 
* Will there be case there will be multiple cli in the same tgz/zip that need to be pull out?
* I can only test manually by changing the url for now because `kubectx` author  is still recommending to use bash version.

## Motivation and Context

Attempt to partially fix #123 

## How Has This Been Tested?

Right now, there's no tgz/zip tool so I manually change the url of `kubectx` to use `tgz` format (hardcode download url to `https://github.com/ahmetb/kubectx/releases/download/v0.9.0/kubectx_v0.9.0_darwin_x86_64.tar.gz` and `NeedDecompress` to ` true` during testing.

image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -4,17 +4,131 @@
 package cmd
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/alexellis/arkade/pkg/env"
 	"github.com/alexellis/arkade/pkg/get"
 	"github.com/spf13/cobra"
 )
+
+// TODO: Untar logic is copied from helm.go. Need to refactor this later on.
+
+// Untar reads the gzip-compressed tar file from r and writes it into dir.
+func Untar(r io.Reader, dir string) error {
+	return untar(r, dir)
+}
+
+func untar(r io.Reader, dir string) (err error) {
+	t0 := time.Now()
+	nFiles := 0
+	madeDir := map[string]bool{}
+	defer func() {
+		td := time.Since(t0)
+		if err == nil {
+			log.Printf("extracted tarball into %s: %d files, %d dirs (%v)", dir, nFiles, len(madeDir), td)
+		} else {
+			log.Printf("error extracting tarball into %s after %d files, %d dirs, %v: %v", dir, nFiles, len(madeDir), td, err)
+		}
+	}()
+	zr, err := gzip.NewReader(r)
+	if err != nil {
+		return fmt.Errorf("requires gzip-compressed body: %v", err)
+	}
+	tr := tar.NewReader(zr)
+	loggedChtimesError := false
+	for {
+		f, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			log.Printf("tar reading error: %v", err)
+			return fmt.Errorf("tar error: %v", err)
+		}
+		if !validRelPath(f.Name) {
+			return fmt.Errorf("tar contained invalid name error %q", f.Name)
+		}
+		baseFile := filepath.Base(f.Name)
+		abs := path.Join(dir, baseFile)
+		fmt.Println(abs, f.Name)
+
+		fi := f.FileInfo()
+		mode := fi.Mode()
+		switch {
+		case mode.IsDir():
+
+			break
+
+		case mode.IsRegular():
+
+			wf, err := os.OpenFile(abs, os.O_RDWR|os.O_CREATE|os.O_TRUNC, mode.Perm())
+			if err != nil {
+				return err
+			}
+			n, err := io.Copy(wf, tr)
+			if closeErr := wf.Close(); closeErr != nil && err == nil {
+				err = closeErr
+			}
+			if err != nil {
+				return fmt.Errorf("error writing to %s: %v", abs, err)
+			}
+			if n != f.Size {
+				return fmt.Errorf("only wrote %d bytes to %s; expected %d", n, abs, f.Size)
+			}
+			modTime := f.ModTime
+			if modTime.After(t0) {
+				// Clamp modtimes at system time. See
+				// golang.org/issue/19062 when clock on
+				// buildlet was behind the gitmirror server
+				// doing the git-archive.
+				modTime = t0
+			}
+			if !modTime.IsZero() {
+				if err := os.Chtimes(abs, modTime, modTime); err != nil && !loggedChtimesError {
+					// benign error. Gerrit doesn't even set the
+					// modtime in these, and we don't end up relying
+					// on it anywhere (the gomote push command relies
+					// on digests only), so this is a little pointless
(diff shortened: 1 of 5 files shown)


cmd/get.go
```diff
@@ -81,14 +84,24 @@ and provides a fast and easy alternative to a package manager.`,
 
 		outFilePath := path.Join(tmp, fileName)
 
-		out, err := os.Create(outFilePath)
-		if err != nil {
-			return err
-		}
-		defer out.Close()
+		if tool.NeedDecompression == true {
```
alexellis, line 87: Could this be a method instead? `IsArchive()`? Look at the URL's suffix?

pkg/get/get.go
```diff
@@ -195,7 +196,8 @@ https://storage.googleapis.com/kubernetes-release/release/{{.Version}}/bin/{{$os
 			Version:     "v0.9.0",
 			URLTemplate: `https://github.com/ahmetb/kubectx/releases/download/{{.Version}}/kubectx`,
```
alexellis, line 197: Theoretically this is now wrong as a URL if you are downloading the archive?

How about leaving this one as it is, and adding helm3 to the list of "get" tools instead?

cmd/get.go
tuananh: Yes, will do. Initially I want to look at the mime type but haven't figure out how to do that with go without third party lib.

pkg/get/get.go
tuananh: helm added 
image

cmd/get.go
```diff
@@ -6,16 +6,23 @@ package cmd
...
 	"strings"
 
 	"github.com/alexellis/arkade/pkg/env"
 	"github.com/alexellis/arkade/pkg/get"
+	"github.com/alexellis/arkade/pkg/helm"
 	"github.com/spf13/cobra"
 )
 
+func IsArchive(url string) bool {
+	return strings.HasSuffix(url, "tar.gz") || strings.HasSuffix(url, "zip")
```
alexellis, line 23: We can't support .zip at this time. I would remove that code.

Also `IsArchive` could be part of the `Tool` type, what do you think?

pkg/get/get_test.go
```diff
@@ -151,3 +151,66 @@ func Test_DownloadWindows(t *testing.T) {
...
+			tool = &target
+			break
+		}
+	}
+
+	got, err := tool.GetURL("mingw64_nt-10.0-18362", arch64bit, tool.Version)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := "https://get.helm.sh/helm-v3.2.4-windows-amd64.zip"
```
alexellis, line 212: Is the un-tar code really going to work with a zip?

alexellis: My gut feeling is: no. However, there is a Go library that can uncompress zips -> https://golang.org/src/archive/zip/example_test.go

alexellis: I may be wrong here, 

```
tar -xvf helm-v3.2.4-windows-amd64.zip 
x windows-amd64/
x windows-amd64/helm.exe
x windows-amd64/README.md
x windows-amd64/LICENSE
```

alexellis: `bsdtar` can uncompress zip archives, but I would think the Golang tar code cannot. In any case using archive/zip for zips would be better suited.

tuananh: @alexellis you're right. it doesn't work with zip file. I add `Unzip` function to deal with that.

Since i don't have a windows machine, I just hardcode a zip url and test . it's able to extract the zip files in the pic below

image

cmd/get.go
tuananh: Updated.

cmd/get.go
```diff
@@ -81,14 +87,45 @@ and provides a fast and easy alternative to a package manager.`,
...
+					return untarErr
+				}
+			} else if strings.HasSuffix(downloadURL, "zip") {
+				buff := bytes.NewBuffer([]byte{})
+				size, err := io.Copy(buff, res.Body)
+				if err != nil {
+					return err
+				}
+				reader := bytes.NewReader(buff.Bytes())
+				zipReader, err := zip.NewReader(reader, size)
```
alexellis, line 109: I think we can improve the API here - for instance, let's pass in the bytes directly to the method and have it create the zip reader via the zip.NewReader call.

pkg/helm/untar.go
```diff
@@ -30,6 +31,80 @@ func Untar(r io.Reader, dir string) error {
...
+		td := time.Since(t0)
+		if err == nil {
+			log.Printf("extracted zip into %s: %d files, %d dirs (%v)", dir, nFiles, len(madeDir), td)
+		} else {
+			log.Printf("error extracting zip into %s after %d files, %d dirs, %v: %v", dir, nFiles, len(madeDir), td, err)
+		}
+	}()
+
+	// Closure to address file descriptors issue with all the deferred .Close() methods
+	extractAndWriteFile := func(f *zip.File) error {
```
alexellis, line 56: Did you copy any of this code from stack overflow or another source like a blog post? If so please quote the URL here for crediting the author.

alexellis:
See #133

PR #161, opened 2020-07-27
Consolidate list of apps and reduce repetition

Waterdrips:
## Description

To add an app previously it was very easy to miss adding it's info to the
info command. This changes the way of adding an app to building a list of
app "objects" and then using this list to get the info message and the
installer command.

I found a few apps which hadn't been added to the info command, probably because it was missed as its yet another thing to add! 

## Motivation and Context

#162 

## How Has This Been Tested?

running install and info on apps.

worked as before

## Types of changes

## Checklist:

- [ ] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide

### DIFF
--- a/cmd/apps/crossplane_app.go
+++ b/cmd/apps/crossplane_app.go
@@ -101,7 +101,7 @@ schedule workloads to any Kubernetes cluster`,
 	return crossplane
 }
 
-const CrossplanInfoMsg = `# Get started by installing a stack for your favorite provider:
+const CrossplaneInfoMsg = `# Get started by installing a stack for your favorite provider:
 * provider-gcp: https://crossplane.io/docs/master/install-crossplane.html#gcp-provider
 * provider-aws: https://crossplane.io/docs/master/install-crossplane.html#aws-provider
 * provider-azure: https://crossplane.io/docs/master/install-crossplane.html#azure-provider
@@ -111,4 +111,4 @@ Learn more about Crossplane: https://crossplaneio.github.io/docs/`
 const crossplaneInstallMsg = `=======================================================================
 = Crossplane has been installed.                                      =
 =======================================================================` +
-	"\n\n" + CrossplanInfoMsg + "\n\n" + pkg.ThanksForUsing
+	"\n\n" + CrossplaneInfoMsg + "\n\n" + pkg.ThanksForUsing
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -6,7 +6,6 @@ package cmd
 import (
 	"fmt"
 
-	"github.com/alexellis/arkade/cmd/apps"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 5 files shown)


Waterdrips:
@alexellis WDYT?

cmd/install.go
```diff
@@ -47,38 +53,55 @@ And to see options for a specific app before installing, run:
...
+	for _, app := range AppList {
+		command.AddCommand(app.Installer())
+	}
 
 	command.AddCommand(MakeInfo())
 
 	return command
 }
+
+var AppList = map[string]ArkadeApp{
```
alexellis, line 65: I'm not keen on this being initialised lazily / at a random time.

What other suggestions do you have for running the code in a deterministic way?

Waterdrips: Could make a func that returns the app list and use that where we need the lists instead of a var in here?

cmd/apps/crossplane_app.go
```diff
@@ -101,7 +101,7 @@ schedule workloads to any Kubernetes cluster`,
 	return crossplane
 }
 
-const CrossplanInfoMsg = `# Get started by installing a stack for your favorite provider:
+const CrossplainInfoMsg = `# Get started by installing a stack for your favorite provider:
```
alexellis, line 104: This is a typo. It's Crossplane

cmd/info.go
```diff
@@ -35,78 +34,16 @@ arkade info --help`,
...
-		case "kube-image-prefetch":
-			msg = apps.KubeImagePrefetchInfoMsg
-		case "registry-creds":
-			msg = apps.RegistryCredsOperatorInfoMsg
-		default:
-			return fmt.Errorf("no info available for app: %s", appName)
+		appList := GetApps()
+		appName := args[0]
+		if _, ok := appList[appName]; !ok {
+			fmt.Printf("No info for app: %s\n", appName)
```
alexellis, line 40: This should return an error as it did before, can you keep the same message?

cmd/install.go
```diff
@@ -45,43 +51,63 @@ And to see options for a specific app before installing, run:
...
+	for _, app := range appList {
+		command.AddCommand(app.Installer())
+	}
 
 	command.AddCommand(MakeInfo())
 
 	return command
 }
+
+func GetApps() map[string]ArkadeApp {
```
alexellis, line 65: This nested format is going to be hard to diff.

Please change it to add each item to the map once initialised.

```
	apps := map[string]ArkadeApp{}
        apps["mongodb"] = NewArkadeApp("mongodb", apps.MakeInstallMongoDB, apps.MongoDBInfoMsg) 
        return apps
```

cmd/install.go
```diff
@@ -45,43 +51,63 @@ And to see options for a specific app before installing, run:
...
+		"grafana":                 NewArkadeApp("grafana", apps.MakeInstallGrafana, apps.GrafanaInfoMsg),
+		"tekton":                  NewArkadeApp("tekton", apps.MakeInstallTekton, apps.TektonInfoMsg),
+		"traefik2":                NewArkadeApp("traefik2", apps.MakeInstallTraefik2, apps.Traefik2InfoMsg),
+		"inlets-operator":         NewArkadeApp("inlets-operator", apps.MakeInstallInletsOperator, apps.InletsOperatorInfoMsg),
+		"nfs-provisioner":         NewArkadeApp("nfs-client-provisioner", apps.MakeInstallNfsProvisioner, apps.NfsClientProvisioneriInfoMsg),
+		"docker-registry":         NewArkadeApp("docker-registry", apps.MakeInstallRegistry, apps.RegistryInfoMsg),
+		"OSM":                     NewArkadeApp("OSM", apps.MakeInstallOSM, apps.OSMInfoMsg),
+		"kube-image-prefetch":     NewArkadeApp("kube-image-prefetch", apps.MakeInstallKubeImagePrefetch, apps.KubeImagePrefetchInfoMsg),
+		"registry-creds":          NewArkadeApp("registry-creds", apps.MakeInstallRegistryCredsOperator, apps.RegistryCredsOperatorInfoMsg),
+		"gitea":                   NewArkadeApp("gitea", apps.MakeInstallGitea, apps.GiteaInfoMsg),
```
alexellis, line 100: I'm seeing the same string duplicated twice on each line - once in the map and one in the function invocation.

alexellis (changes requested):
I think this is moving in the right direction. I've left some comments for your consideration.

cmd/install.go
Waterdrips: yeh i did think about that, ill see if i can work out a better way

Waterdrips:
Thanks, Ill continue to work on it based on the comments

Waterdrips:
Requested changes implemented

alexellis (approved):
Approved

PR #214, opened 2020-09-13
display progress bar when downloading

### LINKED ISSUE #189: Add a progress bar to the download message
It would be useful to have a progress bar when arkade is downloading a binary.

## Expected Behaviour

When a binary is being downloaded, arkade could show the speed, size and percentage of completion. 

## Current Behaviour

Today, nothing is shown. For large binaries, it's hard to tell if the download is working on ot.
## Possible Solution

At okteto, we've had success using https://github.com/cheggaaa/pb/ for this

rberrelleza:
## Description
Display progress information when running `arkade get`

## Motivation and Context

- [x] I have raised an issue to propose this change ([required]

Fixes #189

## How Has This Been Tested?

I tested this by running the following commands, both with progress bar enabled and disabled:

`arkade get civo`
`arkade get terraform`
`arkade get kind`

This covers binary, tar.gz and zip downloads.

## Types of changes

## Checklist:

- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.


rberrelleza:
@alexellis With the current changes, tools that need to be untared or unzipped will not get the progress bar. This is because currently, we display every folder as it gets untared, which doesn't mix well with the progress bar.  What do you think about it?

alexellis:
Can you explain what you mean?

The way I understood it was that the progress bar was for the download, after which it would stop showing progress info?

rberrelleza:
> The way I understood it was that the progress bar was for the download, after which it would stop showing progress info?

Yes, that's the pattern I'm following here.  The challenge with the `tar` files is that it follows a different pattern. Instead of downloading and then processing, is done as a stream. I have some ideas on how to implement this though. I'll update the PR and we can discuss the specifics.

pkg/archive/untar.go
```diff
@@ -61,7 +63,6 @@ func untar(r io.Reader, dir string) (err error) {
 		}
 		baseFile := filepath.Base(f.Name)
 		abs := path.Join(dir, baseFile)
-		fmt.Println(abs, f.Name)
```
rberrelleza, line 64: I'm removing this because it gets in the way of the progress bar (if you print something to stdout before closing the progress bar, it messes up the output). More importantly, it doesn't feel like getting the list of tmp folders extracted is that useful for the end user.

rberrelleza:
@alexellis what do you think of the proposed approach?

pkg/archive/untar.go
alexellis: Isn't the progress bar there for the download of a file? Then when the download is complete, it should stop /complete.

Why is the unzip / untar operation part of the progress bar's scope, if it's only there for the download?

Is it a timing issue? That the progress bar prints some additional output after the download has "finished"?

pkg/get/download.go
```diff
@@ -144,3 +147,21 @@ func copyFile(src, dst string) (int64, error) {
...
+
+	wrapped := withProgressBar(r.Body, int(r.ContentLength))
+	defer wrapped.Close()
+
+	w, err := io.Copy(dst, wrapped)
+	return w, err
+}
+
+func withProgressBar(r io.ReadCloser, length int) io.ReadCloser {
+	if os.Getenv("ARKADE_PROGRESS") == "0" {
```
alexellis, line 161: (I think it should also be a flag)

pkg/archive/untar.go
rberrelleza: There's no "download of a file" for the unzip/untar function in the existing implementation.  `archive.untar` receives the response body (wrapped as a `NopCloser`), and handles all the download internally (https://github.com/alexellis/arkade/blob/master/pkg/archive/untar.go#L44). That's why I went with this approach, where the reader is wrapped in the progress bar. 

If we want to scope the progress bar only to the download step, we could change `get.Download` to first download the file for all cases, and then do the format specific manipulations (e.g. if it's a tar.gz, untar and unzip; for a binary, just move to the final destination).  What approach do you prefer?

alexellis:
Ahh oh I get it now. 

Yes I think it should download to disk and then untar for consistency with the plain downloads.

rberrelleza:
> Ahh oh I get it now.
> 
> Yes I think it should download to disk and then untar for consistency with the plain downloads.

sounds good! last commit has this change.  Let me know if you agree with the changes.

rberrelleza:
@alexellis PR is ready to merge. I applied the changes you requested, and tested it with the different download mechanisms. Let me know if you have any more comments, or if there's any pending work?

pkg/get/download.go
```diff
@@ -119,6 +98,39 @@ func Download(tool *Tool, arch, operatingSystem, version string, downloadMode in
 	return outFilePath, finalName, nil
 }
 
+func downloadFile(downloadURL string, displayProgress bool) (string, error) {
```
alexellis, line 101: I'm a bit conflicted about leaking the progress bar into `downloadFile`, but let's go with it for the time being.

alexellis (approved):
Seems to work well for me -> https://twitter.com/alexellisuk/status/1308331706378858496?s=20

Issue #220, opened 2020-10-10
[App] Sealed Secrets


Waterdrips:
## Expected Behaviour

- App https://github.com/bitnami-labs/sealed-secrets 
~~- CLI tool `kubeseal` added to tools?~~

This would be a useful app to provide for Arkade

alexellis:
Sounds good to me. We already have `arkade get kubeseal` btw

alexellis:
Thanks for picking this up. Have a look at the more recent apps in git history to understand how we are doing them now. Also think about the chart options and whether we need to block on arm (spoiler: we do have to)

evalsocket:
No, today I will start and most probably create a pr forthis issue

alexellis:
Hi @evalsocket is there anything we can help with? We do have Slack available if you want to chat.

PR #239, opened 2020-10-22
Add sealed secret app

evalsocket:
Sealed Secrets App installation

This PR is for installing Sealed Secrets app via Arkade
Issue - #220

### DIFF
--- a/cmd/apps/nginx_app.go
+++ b/cmd/apps/nginx_app.go
@@ -10,9 +10,11 @@ import (
 	"path"
 
 	"github.com/alexellis/arkade/pkg"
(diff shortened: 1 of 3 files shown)


alexellis (changes requested):
Please could you squash these for commits into one? @Waterdrips could you take a look too?

Waterdrips (commented):
This should prpbably be using the new app style like Loki, cert manager and nginx with the installer options and pkg functions.

Also we might want some useful info in info msg (it's empty at the moment?) And the post install message. I'm sure there's some info about using kubeseal or how to check the deployment we could have there

PR #5, opened 2020-02-26
Add helm-operator app

aidun:
Fixes #14 

In alexellis/k3sup#157 @alexellis like to have the helm-operator as app in k3sup.

## Description
The installation of the helm-operator with arkade follows the install progress of the helm-operator. There no custom values at the moment.

## Motivation and Context
alexellis/k3sup#157
- [x] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?
Installed a fresh cluster
```
./k3sup install --ip 192.168.178.75 --user ubuntu
```
Installed the helm-operator
```
./arkade install helm-operator
```
Added a helm deployment
```
kubectl apply -f https://raw.githubusercontent.com/aidun/kubernetes_homelab/master/releases/kube-system/nginx-ingress.yaml
```
Check if helm knows about the deployment
```
helm list -A
```

There should be no effects to the other commands.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #254, opened 2020-10-30
Add helm-operator

tuananh:
Signed-off-by: Tuan Anh Tran <me@tuananh.org>

## Description

Add `helm-operator` in new format. Courtesy the original work from @aidun here https://github.com/alexellis/arkade/pull/5

## Motivation and Context

Related PR: https://github.com/alexellis/arkade/pull/5

- [ ] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

- create a new cluster with `kind` `kind create cluster`
- `./arkade install helm-operator`

output

image

check again with `helm list -A`

```
NAME            NAMESPACE               REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
helm-operator   default                 1               2020-10-30 22:11:10.571254 +0700 +07    deployed        helm-operator-1.2.0     1.2.0  
```

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.
- [ ] All new and existing tests passed.

### DIFF
--- a/cmd/apps/helm_operator_app.go
+++ b/cmd/apps/helm_operator_app.go
@@ -37,7 +37,6 @@ func MakeInstallHelmOperator() *cobra.Command {
 		wait, _ := command.Flags().GetBool("wait")
 		namespace, _ := command.Flags().GetString("namespace")
 		version, _ := command.Flags().GetString("version")
-		helm3, _ := command.Flags().GetBool("helm3")
 		updateRepo, _ := helmOperator.Flags().GetBool("update-repo")
 		kubeConfigPath, _ := command.Flags().GetString("kubeconfig")
 
@@ -68,9 +67,6 @@ func MakeInstallHelmOperator() *cobra.Command {
 		}
 
 		helmRepo := "fluxcd/helm-operator"
-		if !helm3 {
-			helmRepo = "helm-operator/rendered"
-		}
 
 		err = helm.FetchChart(helmRepo, defaultVersion)
 		if err != nil {
--- a/cmd/apps/helm_operator_app.go
+++ b/cmd/apps/helm_operator_app.go
@@ -1,0 +1,141 @@
+// Copyright (c) arkade author(s) 2020. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package apps
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path"
+
+	"github.com/alexellis/arkade/pkg/apps"
+	"github.com/alexellis/arkade/pkg/config"
+	"github.com/alexellis/arkade/pkg/env"
+	"github.com/alexellis/arkade/pkg/helm"
+	"github.com/alexellis/arkade/pkg/k8s"
+	"github.com/alexellis/arkade/pkg/types"
+	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
+)
+
+func MakeInstallHelmOperator() *cobra.Command {
+	var helmOperator = &cobra.Command{
+		Use:          "helm-operator",
+		Short:        "Install helm-operator",
+		Long:         "Install helm-operator",
+		Example:      "arkade install helm-operator --namespace default",
+		SilenceUsage: true,
+	}
+
+	helmOperator.Flags().StringP("namespace", "n", "default", "The namespace used for installation")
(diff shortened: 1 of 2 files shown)


cmd/apps/helm_operator_app.go
```diff
@@ -0,0 +1,141 @@
...
+		installOptions := types.DefaultInstallOptions().
+			WithNamespace(namespace).
+			WithHelmPath(path.Join(userPath, ".helm")).
+			WithHelmRepo(helmRepo).
+			WithHelmURL("https://charts.fluxcd.io").
+			WithOverrides(overrides).
+			WithWait(wait).
+			WithHelmUpdateRepo(updateRepo)
+
+		if command.Flags().Changed("kubeconfig") {
```
Waterdrips, line 97: We should always be getting and setting the kubeconfig flag value and setting it in the options, not only when its changed. 

always grab the value at the start
```go
kubeConfigPath, _ := command.Flags().GetString("kubeconfig")
```
then add the below to the optionas

```go 
			WithKubeconfigPath(kubeConfigPath)

```

Waterdrips (changes requested):

cmd/apps/helm_operator_app.go
```diff
@@ -0,0 +1,137 @@
...
+		Use:          "helm-operator",
+		Short:        "Install helm-operator",
+		Long:         "Install helm-operator",
+		Example:      "arkade install helm-operator --namespace default",
+		SilenceUsage: true,
+	}
+
+	helmOperator.Flags().StringP("namespace", "n", "default", "The namespace used for installation")
+	helmOperator.Flags().Bool("update-repo", true, "Update the helm repo")
+	helmOperator.Flags().Bool("helm3", true, "Use helm3, if set to false uses helm2")
```
Waterdrips, line 33: just spotted helm2. We no longer support helm2 in arkade, can this be taken out

Waterdrips (changes requested):

cmd/apps/helm_operator_app.go
tuananh: Removed

cmd/apps/helm_operator_app.go
tuananh: Done

cmd/apps/helm_operator_app.go
```diff
@@ -0,0 +1,133 @@
...
```
Waterdrips, line 33: need to remove the --helm3 flag still, the code that uses it has gone.

cmd/apps/helm_operator_app.go
```diff
@@ -0,0 +1,133 @@
...
+
+		helmRepo := "fluxcd/helm-operator"
+
+		err = helm.FetchChart(helmRepo, defaultVersion)
+		if err != nil {
+			return err
+		}
+
+		overrides := map[string]string{}
+		overrides["helm.versions"] = "v3"
```
Waterdrips, line 77: This flag, looking at the helm operator readme, states if the helm operator install should allow helm2&3 or just helm3 charts to be installed (by the running operator)

We should either leave it out, or have a flag like `--helm3-only` that's false by default, and not setting this override. 

https://github.com/fluxcd/helm-operator/tree/master/chart/helm-operator#installation

Waterdrips (changes requested):

cmd/apps/helm_operator_app.go
tuananh: not sure what I should do here since arkade only supports helm3 from now?

Waterdrips: good point

The readme states 

```
Helm versions supported by this operator instance, if v2 is specified then Tiller is required
```

We should probably just set it to v3, then later on (after merging a user's --set flags) check if they have put something that contains v2 and then print something along the lines of `you have specified helm2 support for the operator, please ensure tiller is installed on your cluster` 

Or you could look at doing a check for the tiller pod before trying to install? 

Im not sure what happens if you try to use `v2` without tiller pre-installed. can you check and come up with a plan?

Shikachuu:
@tuananh, are you still intrested?

Shikachuu:
I think we can assume that this PR has been abandoned by the author.
I'll close this for now, but feel free to revisit this in a new PR!

PR #258, opened 2020-10-31
Fix Kubeconfig loading as per #15

### LINKED ISSUE #15: Fix --kubeconfig flag priority
The `kubeconfig` flag on the `app install` command does not affect the actual kubernetes command execution. The kubernetes command is executed against the default kubeconfig file, not the one passed in via flags.

## Expected Behaviour

Passing the `kubeconfig` flag on the `app install` command should use that kubeconfig file when installing the app.

## Current Behaviour

The `kubeconfig` variable is only written out, but not utilized:
https://github.com/alexellis/k3sup/blob/master/pkg/cmd/openfaas_app.go#L42

## Possible Solution

Since the `kubernetes_exec.go` file is also in the `cmd` package, it's possible for the `kubectl` and `kubectlTask` functions to look for changes on the flag to get that passed in value. Much the same as the above pasted code from one of the apps.

## Steps to Reproduce (for bugs)

1. `k3sup app install openfaas --kubeconfig any/path/to/file`
2. Output will say something like: "Using kubeconfig {the path from above}"
3. Run `kubectl get pods --all-namespaces --kubeconfig /same/path/above`
4. Observe the app was not installed
5. Run `kubectl get pods --all-namespaces` (without the kubeconfig)
6. See the application was installed on the incorrect cluster

## Context

I have several k8s cluster set up in various environments for testing. Swapping the config files, or env var is annoying, and really time consuming if I forget before I run the install command.

## Your Environment

* What OS or type or VM are you using? Where is it hosted?
Linux/ARM
Desktop (microk8s), GCP, DigitalOcean, Raspberry Pi 3+
* Operating System and version (e.g. Linux, Windows, MacOS):
Linux

Waterdrips:
## Description

Kubeconfig flag was not honoured if set.
This commit checks the flag, then os.env then the default kueconfig
location.

Tested by installing a handful of apps (new and old style) with and
without kubeconfig flag and os.env
All installed ok

## Motivation and Context

Fixes #15

## How Has This Been Tested?

installing and building locally with various apps with and without the kubeconfig flag
## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [ ] I have signed-off my commits with `git commit -s`
- [ ] All new and existing tests passed.

### DIFF
--- a/cmd/apps/argocd_app.go
+++ b/cmd/apps/argocd_app.go
@@ -25,13 +25,10 @@ func MakeInstallArgoCD() *cobra.Command {
 	}
 
 	command.RunE = func(command *cobra.Command, args []string) error {
-		kubeConfigPath := config.GetDefaultKubeconfig()
-
-		if command.Flags().Changed("kubeconfig") {
-			kubeConfigPath, _ = command.Flags().GetString("kubeconfig")
+		kubeConfigPath, _ := command.Flags().GetString("kubeconfig")
+		if err := config.SetKubeconfig(kubeConfigPath); err != nil {
+			return err
 		}
-
-		fmt.Printf("Using kubeconfig: %s\n", kubeConfigPath)
 
 		arch := k8s.GetNodeArchitecture()
 		fmt.Printf("Node architecture: %q\n", arch)
--- a/cmd/apps/openfaas_app.go
+++ b/cmd/apps/openfaas_app.go
@@ -54,17 +54,17 @@ func MakeInstallOpenFaaS() *cobra.Command {
 	openfaas.Flags().StringArray("set", []string{}, "Use custom flags or override existing flags \n(example --set=gateway.replicas=2)")
 
 	openfaas.RunE = func(command *cobra.Command, args []string) error {
-		kubeConfigPath := config.GetDefaultKubeconfig()
 		wait, err := command.Flags().GetBool("wait")
 		if err != nil {
 			return err
 		}
-		if command.Flags().Changed("kubeconfig") {
-			kubeConfigPath, _ = command.Flags().GetString("kubeconfig")
+
+		kubeConfigPath, _ := command.Flags().GetString("kubeconfig")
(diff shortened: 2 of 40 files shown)


cmd/apps/argocd_app.go
```diff
@@ -25,14 +25,11 @@ func MakeInstallArgoCD() *cobra.Command {
 	}
 
 	command.RunE = func(command *cobra.Command, args []string) error {
-		kubeConfigPath := config.GetDefaultKubeconfig()
-
-		if command.Flags().Changed("kubeconfig") {
-			kubeConfigPath, _ = command.Flags().GetString("kubeconfig")
+		kubeConfigPath, _ := command.Flags().GetString("kubeconfig")
```
alexellis, line 28: Wonder if this could be moved into the install command?

alexellis: i.e. as a global flag

cmd/apps/openfaas_app.go
```diff
@@ -54,17 +54,17 @@ func MakeInstallOpenFaaS() *cobra.Command {
...
-			kubeConfigPath, _ = command.Flags().GetString("kubeconfig")
-		}
 
+		kubeConfigPath, _ := command.Flags().GetString("kubeconfig")
+		if err := config.SetKubeconfig(kubeConfigPath); err != nil {
+			return err
+		}
 		fmt.Printf("Using kubeconfig: %s\n", kubeConfigPath)
-
+		os.Setenv("KUBECONFIG", kubeConfigPath)
```
wilsonianb, line 67: If `KUBECONFIG` is defined but the `--kubeconfig` flag is not set, this appears to overwrite `KUBECONFIG` with nothing causing a fallback to `$HOME/.kube/config`.

Issue #271, opened 2020-11-09
Cannot give additional helm options to cert-manager


fabn:
I want to enable the [`kubernetes.io/tls-acme: "true"`](https://cert-manager.io/docs/usage/ingress/#optional-configuration) in my cert manager deployment and I tried with

```
arkade install cert-manager \
--set=ingressShim.defaultIssuerName=letsencrypt-prod \
--set=ingressShim.defaultIssuerKind=ClusterIssuer
```

## Expected Behaviour

Helm chart should be installed with those options set

Also a dedicated flag such as `--acme` would be nice

## Current Behaviour

```
$ arkade install cert-manager --set=ingressShim.defaultIssuerName=letsencrypt-prod --set=ingressShim.defaultIssuerKind=ClusterIssuer
Error: unknown flag: --set
```

## Possible Solution

Currently I'm solving this with this command after a successfull install

```
helm upgrade --install cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --reuse-values \
  --set=ingressShim.defaultIssuerName=letsencrypt-prod \
  --set=ingressShim.defaultIssuerKind=ClusterIssuer \
  --set ingressShim.defaultIssuerGroup=cert-manager.io
```

Long term solution would be to allow [`--set`](https://github.com/alexellis/arkade/blob/master/cmd/apps/openfaas_app.go#L57) in cert-manager install command. Or even better in all install subcommands. 

From current Readme I understand that `--set` is always allowed, but looking at source code that's not the case.

## Your Environment

* What Kubernetes distribution are you using?

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:49:28Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:41:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

* What arkade version is this?

```
$ arkade version
Version: 0.6.18
Git Commit: 518f66112256610965d8bd28041894a733af0c4c
```

[...]

PR #272, opened 2020-11-09
Allow acme boolean flag to enable lego behavior with default issuer

fabn:
Signed-off-by: Fabio Napoleoni <f.napoleoni@gmail.com>

## Description

Implement a `--acme` flag for cert-manager install subcommand that automates helm values 

```
--set=ingressShim.defaultIssuerName=letsencrypt-prod \
--set=ingressShim.defaultIssuerKind=ClusterIssuer
```

and create a global `ClusterIssuer` that is used to generate letsencrypt certificates when an ingress is annotated with `kubernetes.io/tls-acme: "true"`

Let me know if you're interested and if I should change something to make this merged.

## Motivation and Context

As explained in #271 it would be useful to have an app specific flag that implements [`kubernetes.io/tls-acme: "true"`](https://cert-manager.io/docs/usage/ingress/#optional-configuration) configuration.

I think is in the motivation of this project to have some shorthand to automate per cluster tasks that are tedious and error prone like the one above.

Also the implementation is completely optin.

## How Has This Been Tested?

I built the binary locally and tested on my own cluster, I'm not so skilled with go (actually my first lines in go) thus I'm still not able to implement an automated test. If you give me some hint on how to do that I'll be glad to try.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/apps/certmanager_app.go
+++ b/cmd/apps/certmanager_app.go
@@ -4,14 +4,26 @@
 package apps
 
 import (
+	"bytes"
 	"fmt"
+	"log"
+
+	"text/template"
 
 	"github.com/alexellis/arkade/pkg"
 	"github.com/alexellis/arkade/pkg/apps"
+	"github.com/alexellis/arkade/pkg/k8s"
 	"github.com/alexellis/arkade/pkg/types"
 	"github.com/spf13/cobra"
 	"golang.org/x/mod/semver"
 )
+
+type IssuerInput struct {
+	CertmanagerEmail string
+	IngressClass     string
+	IssuerName       string
+	IssuerAPI        string
+}
 
 func MakeInstallCertManager() *cobra.Command {
 	var certManager = &cobra.Command{
@@ -25,6 +37,12 @@ func MakeInstallCertManager() *cobra.Command {
 	certManager.Flags().StringP("namespace", "n", "cert-manager", "The namespace to install cert-manager")
 	certManager.Flags().StringP("version", "v", "v1.5.4", "The version of cert-manager to install, has to be >= v1.0.0")
 	certManager.Flags().Bool("update-repo", true, "Update the helm repo")
+	// Acme options
+	certManager.Flags().Bool("acme", false, "Enable automated usage of kubernetes.io/tls-acme: \"true\" annotation (implies wait)")
+	certManager.Flags().Bool("staging", false, "set --staging to true to generate a staging Letsencrypt cluster issuer")
+	certManager.Flags().StringP("email", "e", "", "Letsencrypt Email")
+	certManager.Flags().String("ingress-class", "nginx", "Ingress class to be used such as nginx or traefik")
+	// Override flags
 	certManager.Flags().StringArray("set", []string{}, "Use custom flags or override existing flags \n(example --set key=value)")
 
 	certManager.RunE = func(command *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


Waterdrips:
Hey, we like to discuss changes like this on issues, I know we had one to discuss the missing `--set` flag from cert-manager app. I think you also proposed adding the functionality in this PR but it didnt really get addressed.

I'm not sure of the implementation of this feature, we have some apps creating LE issuers (openfaas-ingress and docker-registry-ingress). They can install a staging or production issuer with different ingressClass annotations (traefik for example) and they have templated the YAML used to create the issuer, this lets us do the above modifications as well as not require using yaml from the internet (which may change between cli versions). 

Additionally the `--acme` flag you have added is true by default, which is either a breaking change (if someone has an issuer called that already) or will start to install cluster-scoped resources into people's clusters without them knowing.

fabn:
Good points, I'll try to answer:

* true by default was obviously a mistake, my proposed change should be optin. I'll change it
* for the yaml part I completely agree that that the proposed approach is not so clean and my first idea was to embed some kind of  yaml template in the command, also because that will allow further customization (e.g. to register a letsencrypt email address in issuer) but as I told you my go skills are pretty limited. If you point me to the right direction I'll change it to use a template instead of downloading the issuer template from git repo.

fabn:
I've update PR with suggested changes, let me know if they aree good.

fabn:
@Waterdrips feedback?

Waterdrips:
Looks good, can we add a --staging flag like we have for the other apps that create issuers? 

Good for testing and not using your rate limit up.

fabn:
@Waterdrips done. I have a test failure but is unrelated to my changes. I think is because of openfaas new releasees.

fabn:
@Waterdrips up

Waterdrips (changes requested):
2 things,
1. can you update the tests to look for the new faas-cli verison.
2. the issuer secret needs to be named after the correct issuer (stg/prod)

Also, the commits need squashing down into 1 commit which should have a good commit message https://chris.beams.io/posts/git-commit/

fabn:
Done, please check it out.

cmd/apps/certmanager_app.go
```diff
@@ -95,6 +111,13 @@ func MakeInstallCertManager() *cobra.Command {
 			overrides["installCRDs"] = "true"
 		}
 
+		// Allow usage of lego behavior, see https://cert-manager.io/docs/usage/ingress/#optional-configuration
+		useAcme, _ := certManager.Flags().GetBool("acme")
+		if useAcme {
+			overrides["ingressShim.defaultIssuerName"] = "letsencrypt-production-issuer"
```
Waterdrips, line 117: I think this needs the same staging logix as the issuer as here, https://github.com/alexellis/arkade/pull/272/files#diff-90097b757fed057b84f0204d467ed1708d81bdf17fc9e0188fd9c22fcc04b577R219-R222

So the sataging issuer name is used if --staging is set?

Waterdrips (changes requested):
one small change for the overrides map, need to use staging if --staging is set

cmd/apps/certmanager_app.go
fabn: You're right, that's needed as well.

I'll change it

fabn: Done.

fabn:
Up

Shikachuu:
@fabn are you still intrested, to move the PR forward?

fabn:
I used this fork on more than one cluster and it works as is. What should I do to get it merged?

cmd/apps/certmanager_app.go
```diff
@@ -129,7 +173,94 @@ func MakeInstallCertManager() *cobra.Command {
...
+	if templateErr != nil {
+		log.Print("Unable to install the application. Could not build the templated yaml file for the resources")
+		return templateErr
+	}
+
+	tempFile, tempFileErr := writeTempFile(yamlBytes, "temp_registry_ingress.yaml")
+	if tempFileErr != nil {
+		log.Print("Unable to save generated yaml file into the temporary directory")
+		return tempFileErr
+	}
```
Shikachuu, line 197: Imo it would be more reasonable to `fmt.Errorf` the log line, like in line 206 or every other application.

Shikachuu (changes requested):
Everything except the log lines looks good to me. (Also thats just a code style issue tho'.)

cmd/apps/certmanager_app.go
fabn: Done

Shikachuu:
@fabn can you rebase your commits and sign-off your commit please?

fabn:
> @fabn can you rebase your commits and sign-off your commit please?

Done

fabn:
Also fixed some compile error when rebased.

Shikachuu:
LGTM

Shikachuu (approved):

PR #297, opened 2020-12-15
Make inlets-operator compatible with various K8s versions

alexellis:
Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

## Description

Make inlets-operator compatible with various K8s versions

## Motivation and Context

This checks the k8s version before using a dryrun=client
flag.

@Waterdrips reported an issue with an older kubectl version (1.17.x)

## How Has This Been Tested?

Tested with arkade get kubectl v1.17.3 and v1.19.5 - in each
case arkade "did the right thing"

## Types of changes

### DIFF
--- a/cmd/apps/inletsoperator_app.go
+++ b/cmd/apps/inletsoperator_app.go
@@ -5,12 +5,14 @@ package apps
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Waterdrips (approved):
Lgtm

PR #302, opened 2020-12-31
Fixes #301 Updated HashiCorp products to latest version

### LINKED ISSUE #301: HashiCorp products are not using the latest versions as the default.
## Expected Behaviour

When executing the following command I expect version 0.14.3 of Terraform to be the version that is installed, not 0.13.1. The same thing happens when installing Packer.

```
arkade get terraform
```
## Current Behaviour

By default Arkade installs older versions of both Packer and Terraform.

## Possible Solution

Update the default values for version to the latest release.

## Context

I just recently started using Arkade to installing tools and those tools are older versions than I typically use. I was able to install by passing the version flag. That just may not be what is expected by most users as they wouldn't know what version gets installed by default with reading the source.

0.6.5

phillipsj:
## Description

The default for Packer and Terraform were not at the latest versions.
This update brings both of those to the newest version by default.
Vagrant is at the current latest version. Linked to issue #302 

## Motivation and Context

## How Has This Been Tested?

I ran the make command and everything compiled and all tests passed. I just updated the version for the Terraform and Packer tools.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/get/tools.go
+++ b/pkg/get/tools.go
@@ -456,13 +456,13 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Name}}
 https://github.com/{{.Owner}}/{{.Repo}}/releases/download/v{{.Version}}/{{.Name}}-{{.Version}}-{{$osStr}}-{{$archStr}}.{{$extStr}}`,
 		})
 
-	// https://releases.hashicorp.com/terraform/0.13.1/terraform_0.13.1_linux_amd64.zip
+	// https://releases.hashicorp.com/terraform/0.14.3/terraform_0.14.3_linux_amd64.zip
 	tools = append(tools,
 		Tool{
 			Owner:   "hashicorp",
 			Repo:    "terraform",
 			Name:    "terraform",
-			Version: "0.13.1",
+			Version: "0.14.3",
 			URLTemplate: `{{$arch := .Arch}}
 
 {{- if eq .Arch "x86_64" -}}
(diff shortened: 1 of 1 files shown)


pkg/get/tools.go
```diff
@@ -456,13 +456,13 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Name}}
 https://github.com/{{.Owner}}/{{.Repo}}/releases/download/v{{.Version}}/{{.Name}}-{{.Version}}-{{$osStr}}-{{$archStr}}.{{$extStr}}`,
 		})
 
-	// https://releases.hashicorp.com/terraform/0.13.1/terraform_0.13.1_linux_amd64.zip
+	// https://releases.hashicorp.com/terraform/0.14.3/terraform_0.14.3_linux_amd64.zip
```
alexellis, line 459: @jsiebens / @Waterdrips does this change break any of our faasd tutorials?

alexellis:
Thanks for the PR!

pkg/get/tools.go
Waterdrips: The tf 0.14 upgrade docs don't mention any breaking changes to config syntax. Might be worth checking them with 0.14 though.

phillipsj:
If you could point me to what you are reviewing I would be happy to help. I
use Terraform almost every day.

alexellis:
We have a bunch in the openfaas/faasd repo and a post on openfaas.com/blog that relies on the user having 0.12.0.

Issue #436, opened 2021-06-20
Kubectl install fails with 404 because of incorrect url


LucasRoesler:
The github.com/openfaas/certifier CI is failing because arkade fails to install kubectl

## Expected Behaviour

[This script](https://github.com/openfaas/certifier/blob/master/contrib/get_tools.sh) should work, as it did previously

```sh
KUBE_VERSION=v1.18.8

echo ">>> Installing kubectl $KUBE_VERSION"
arkade get kubectl --version $KUBE_VERSION
sudo mv $HOME/.arkade/bin/kubectl /usr/local/bin/
```

## Current Behaviour

```sh
>>> Installing kubectl v1.18.8
Downloading kubectl
https://storage.googleapis.com/kubernetes-release/release/vv1.18.8/bin/linux/amd64/kubectl
Error: Process completed with exit code 1.
```

The url is incorrect, it contains two `v` in the path: `vv1.18.8`

## Possible Solution

Ideally the installation should be able to handle specifying the version with or without a `v` and still build the correct URL

## Steps to Reproduce (for bugs)

1. Run `arkade get kubectl --version v1.18.8`

## Your Environment

* What Kubernetes distribution are you using?
Not applicable

* What arkade version is this?

```
arkade version
```

```
Version: 0.7.19
Git Commit: a6a37d06ae25663b03f6279c3aaf7b01d3505eca
```

## Are you a GitHub Sponsor (Yes/No?)

Check at: https://github.com/sponsors/alexellis
- [x] Yes
- [ ] No

alexellis:
From my perspective, this is still working as designed.

```
alex@am1 ~ % arkade get kubectl
Downloading kubectl
https://storage.googleapis.com/kubernetes-release/release/v1.20.0/bin/darwin/amd64/kubectl
44.09 MiB / 44.09 MiB [--------------------------------------------------------------------------------------------------------] 100.00%
Tool written to: /Users/alex/.arkade/bin/kubectl

# Add (kubectl) to your PATH variable
export PATH=$PATH:$HOME/.arkade/bin/

# Test the binary:
/Users/alex/.arkade/bin/kubectl

# Or install with:
sudo mv /Users/alex/.arkade/bin/kubectl /usr/local/bin/

alex@am1 ~ % arkade get kubectl --version 1.20.0
Downloading kubectl
https://storage.googleapis.com/kubernetes-release/release/v1.20.0/bin/darwin/amd64/kubectl
[...]

PR #430, opened 2021-06-08
Create test table for kubectl tool

dirien:
Fixes: https://github.com/alexellis/arkade/issues/423

Signed-off-by: Engin Diri <engin.diri@mail.schwarz>

## Motivation and Context

- [x] I have raised an issue to propose this change ([required] (https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

https://github.com/alexellis/arkade/issues/423

## How Has This Been Tested?

I run the tests in my IDE
image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

- [ ] I have tested this on arm, or have added code to prevent deployment

PR #437, opened 2021-06-21
Fix kubectl incorrect download URL version

yankeexe:
Specify `v` on the version element of Tool struct for kubectl.

Addresses #436 

Signed-off-by: Yankee Maharjan <yankee.exe@gmail.com>

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/pkg/get/tools.go
+++ b/pkg/get/tools.go
@@ -101,7 +101,7 @@ helmfile_{{$os}}_{{$arch}}{{$ext}}`,
 			Owner:       "kubernetes",
 			Repo:        "kubernetes",
 			Name:        "kubectl",
-			Version:     "1.20.0",
+			Version:     "v1.20.0",
 			Description: "Run commands against Kubernetes clusters",
 			URLTemplate: `{{$arch := "arm"}}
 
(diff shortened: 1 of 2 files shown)


pkg/get/tools.go
```diff
@@ -121,7 +121,7 @@ helmfile_{{$os}}_{{$arch}}{{$ext}}`,
 {{$os = "windows"}}
 {{- end -}}
 
-https://storage.googleapis.com/kubernetes-release/release/v{{.Version}}/bin/{{$os}}/{{$arch}}/kubectl{{$ext}}`})
+https://storage.googleapis.com/kubernetes-release/release/{{.Version}}/bin/{{$os}}/{{$arch}}/kubectl{{$ext}}`})
```
alexellis, line 124: I can't see the difference. We had:

v + 1.20.0

Then the change moves the v to the Version string, so it still prints v1.20.0

alexellis: Here's an example of what I mean for `doctl`:

https://github.com/alexellis/arkade/blob/master/pkg/get/tools.go#L406

With the version passed in without a `v` prefix

https://github.com/alexellis/arkade/blob/master/pkg/get/tools.go#L382

Perhaps we could have a `v` where that corresponds to the actual GitHub release, and then a Go templates function which can add the prefix if missing, but which won't duplicate it?

alexellis: OK I can see that this was broken in #430 in a PR by @dirien 

https://github.com/alexellis/arkade/pull/430/files

This is on me for not picking it up. Sorry about that.

dirien:
Sorry, i was not sure what the prefered way is.  the version in the version tag of arkade or in the url... Sorry for the trouble...

PR #447, opened 2021-07-06
Use text/template instead of html/template for binary url

soider:
Signed-off-by: Mikhail Sakhnov <msakhnov@mirantis.com>

## Description
The html/template claims to be template renderer engine which outputs html
It sanitizes outputs the same way as html by adding escaping

That breaks final URL if version or tool name or any other used field has different special signs in it

## Motivation and Context
If there are + signs or any other escaping signs in the version or name, the final url is broken.

## How Has This Been Tested?
By running existed test suite and manually downloading few tools

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide

### DIFF
--- a/pkg/get/tools.go
+++ b/pkg/get/tools.go
@@ -1,6 +1,8 @@
 package get
 
-import "strings"
+import (
+	"strings"
+)
 
 func (t Tools) Len() int { return len(t) }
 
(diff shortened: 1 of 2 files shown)


alexellis:
Hi @soider as part of our contributing process, can you raise an issue for this please and give a specific use-case that shows a template failing?

Thanks,

Alex

alexellis:
In the meantime I'll approve the CI to run.

pkg/get/tools.go
```diff
@@ -1310,5 +1313,69 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Name}}
...
+			{{- end -}}
+			{{ if HasPrefix .OS "ming" -}}
+			{{$ext = ".exe"}}
+			{{- end -}}
+			{{.Name}}-{{.Version}}-{{$arch}}{{$ext}}
+			`,
+		},
+	)
+	// k0s binary is available only for linux and windows
+	if runtime.GOOS != "darwin" {
```
alexellis, line 1339: We have code to handle this, so would just remove that if guard.

alexellis (changes requested):
This is shaping up very nicely. Could you just take a look at the note I left for you and amend that last commit and force push after?

alexellis:
@yankeexe we have a PR for K0s already here, so we may want to close #454

alexellis:
@soider please see the error: https://github.com/alexellis/arkade/pull/447/checks?check_run_id=3056341788

alexellis:
FYI we have a rebase conflict on this now too. Could you fix? https://www.youtube.com/watch?v=8j0H6urZ-bU

soider:
I resolved the conflict and dropped the check for darwin OS

soider:
@alexellis as I mentioned, I dropped the OS check in the tools list, though I still able to download binary on darwin (while the binary doesn't support darwin)

alexellis:
We have another rebase issue, sorry about that - due to merging #457 

Once resolved, I'll merge this and then make a recommendation about Darwin. We can detect the OS during the string building template, so that might be a good place to add extra text that would make it fail to match a valid binary.

soider:
@alexellis rebased on master branch

alexellis:
Merged, thank you for the patience.

soider:
@alexellis and thank you for the hints as well!

alexellis:
Want to try that suffix that's normally empty but breaks the URL delicately for Darwin?

soider:
@alexellis I am not sure I understand what are you about. Could you explain a little further?

alexellis:
Didn't you tell me that k0s downloads in error on Darwin?

Break its URL on purpose so that it gives the error saying that it's unavailable.

soider:
Oh, yeah, got you. Good point actually, if the k0s changed the url scheme, I'd re-iterate the changes introduced here.

Issue #459, opened 2021-07-14
Ingress is moving in K8s 1.22, we need to update our Ingress generators


alexellis:
## Expected Behaviour

When using K8s 1.22 and newer, use the new namespace for ingress, for older use the deprecated API (`apiVersion: extensions/v1beta1`)

## Current Behaviour

Only uses the deprecated API

## Possible Solution

Update openfaas-ingress and any other code that generates ingress definitions the same way.

## Steps to Reproduce (for bugs)

1. Start minikube or kind with K8s 1.22
2. Try to install openfaas, then openfaas-ingress
3. It should fail

## Context

Blocks users from using these features.

https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes

https://github.com/alexellis/arkade/blob/master/cmd/apps/openfaas_ingress_app.go#L288

developer-guy:
I can take this one if no one is interested in it 🙋🏻‍♂️

alexellis:
Great, this is time sensitive, I expect users to start complaining next week when 1.22 gets released to Minikube etc.

Can you start with a PR for the openfaas-ingress app?

The Kubernetes version is readable from the code already, and we may already use it somewhere.

Alex

ocpu:
After #462 deprecated ingress template configurations are still generated for docker-registry-ingress app with Kubernetes version 1.22 as there is a missing check to switch the templates in [`registry_ingress_app.go` at line 111](https://github.com/alexellis/arkade/blob/ae978e84af34ad55f72b807529bafcc5d774d26c/cmd/apps/registry_ingress_app.go#L111). The check is there in the `openfaas_ingress_app.go` file so that one works.

References:
- [(master branch) `registry_ingress_app.go#L111`](https://github.com/alexellis/arkade/blob/master/cmd/apps/registry_ingress_app.go#L111)
- [(master branch) `openfaas_ingress_app.go#L201`](https://github.com/alexellis/arkade/blob/master/cmd/apps/openfaas_ingress_app.go#L201)
- [(PR commit) `registry_ingress_app.go#L111`](https://github.com/alexellis/arkade/blob/ae978e84af34ad55f72b807529bafcc5d774d26c/cmd/apps/registry_ingress_app.go#L111)
[...]

PR #462, opened 2021-07-17
Fix ingress support in k8s 1.22

LucasRoesler:
## Description
Add utility to retrieve the list of supported apis from the current k8s
context.

Use this to pass a flag to the ingress creation for the openfaas ingress
app. This allows the app to install with the newer networking.k8s.io/v1
api instead of the deprecated extensions/v1 api that will be removed in
k8s 1.22+ (July 2021)

## Motivation and Context

#459 

## How Has This Been Tested?

New unit test

- [x] Yes
- [ ] No

## Types of changes

## Checklist:

### DIFF
--- a/cmd/apps/registry_ingress_app.go
+++ b/cmd/apps/registry_ingress_app.go
@@ -33,8 +33,8 @@ func MakeInstallRegistryIngress() *cobra.Command {
 	var registryIngress = &cobra.Command{
 		Use:   "docker-registry-ingress",
 		Short: "Install registry ingress with TLS",
-		Long: `Install registry ingress. Requires cert-manager 0.11.0 or higher installation 
-in the cluster. Please set --domain to your custom domain and set --email 
+		Long: `Install registry ingress. Requires cert-manager 0.11.0 or higher installation
+in the cluster. Please set --domain to your custom domain and set --email
 to your email - this email is used by letsencrypt for domain expiry etc.`,
 		Example:      `  arkade install registry-ingress --domain registry.example.com --email openfaas@example.com`,
 		SilenceUsage: true,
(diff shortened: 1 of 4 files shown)


pkg/k8s/kubernetes_exec.go
```diff
@@ -13,13 +14,38 @@ import (
...
 func GetNodeArchitecture() string {
 	res, _ := KubectlTask("get", "nodes", `--output`, `jsonpath={range $.items[0]}{.status.nodeInfo.architecture}`)
 
 	arch := strings.TrimSpace(string(res.Stdout))
 
 	return arch
 }
+
+// GetCapabilities returns the supported API versions on the server
+func GetCapabilities() (Capabilities, error) {
```
alexellis, line 29: Interesting approach. I quite like that 💪

cmd/apps/registry_ingress_app.go
```diff
@@ -205,3 +213,48 @@ spec:
     - http01:
         ingress:
           class: {{.IngressClass}}`
+
+// Ingress in networking.k8s.io/v1 was added in k8s 1.19+
+// this includes the pathType change added in 1.18
+var registryIngressNetworkingYamlTemplate = `
```
alexellis, line 219: Reading this, I wonder whether you concluded that is it easier for us to maintain two templates and phase out the older one in due course than use more conditionals?

alexellis (commented):
Nice work. I'm leaving a couple of comments because I'd like to know more.

cmd/apps/registry_ingress_app.go
LucasRoesler: I picked separate templates because i thought it would be more maintainable for myself and other people. It was just easier to read, the `if` statements in the template are a bit messy and as you point out, we can probably just _drop_ the the old template in a year or so and deleting a template is easier than undoing the if statements, imo.

pkg/k8s/kubernetes_exec.go
LucasRoesler: I was inspired by a similar feature in helm and I thought this was cleaner than checking semver. It is directly answering the question "does this cluster support this api" instead of depending on us reading documentation and changelogs correctly to write the correct semver requirement and then documenting that decision in our own code.

alexellis:
Thanks for the reply. Merging now

PR #519, opened 2021-09-11
Add --dry-run feature for downloading tools

### LINKED ISSUE #518: Add --dry-run feature to arkade get
Before downloading tools via `arkade get` it would be great to view what version of tool the user would be downloading upon running `arkade get <tool-name>` .

## Expected Behaviour

The user being able to run `--dry-run` flag with a tool name to view what version of the tool is being downloaded, where the binary will be stored. In general what the download would look like without actually downloading the tool.

## Current Behaviour

User comes and runs `arkade get <tool-name>` without being sure of what version is being downloaded, where it will be stored, and what to do upon completion of the download.

Check at https://github.com/sponsors/alexellis

- [ ] Yes
- [x] No

## Possible Solution

Add `--dry-run` flag to `arkade get` to simulate the tool download with all the logs and instructions for the user.

yankeexe:
Signed-off-by: Yankee Maharjan <yankee.exe@gmail.com>

## Description

Allows the user to run a dry run of tool download to view what version of the tool is being downloaded, and what up version is available. In general what the download would look like without actually downloading the tool.

## Motivation and Context

## How Has This Been Tested?

dry_run arkade

- [ ] Yes
- [x] No

## Types of changes

## Documentation

- [x] I have updated the list of tools in README.md if (required) with `./arkade get -o markdown`

## Checklist:

Closes #518

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,10 @@ arkade get APP@VERSION
 
 # Download multiple tools at once
 arkade get APP1 APP2 APP3
-
 arkade get APP1 APP2@VERSION APP3
+
+# Dry Run
+arkade get APP@VERSION --dry-run
 ```
 
 |       TOOL       |                                                                DESCRIPTION                                                                |
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -36,6 +36,7 @@ and provides a fast and easy alternative to a package manager.`,
   arkade get terraform --version=0.12.0
   arkade get kubectl --progress=false
   arkade get kubectl@v1.19.3
+  arkade get kubectl --dry-run
 
   # Get a complete list of CLIs to download:
   arkade get`,
@@ -48,10 +49,20 @@ and provides a fast and easy alternative to a package manager.`,
 	command.Flags().StringP("output", "o", "", "Output format of the list of tools (table/markdown/list)")
 	command.Flags().Bool("stash", true, "When set to true, stash binary in HOME/.arkade/bin/, otherwise store in /tmp/")
 	command.Flags().StringP("version", "v", "", "Download a specific version")
+	command.Flags().Bool("dry-run", false, "When set to true, performs a dry run of binary download and install location")
 
 	command.RunE = func(cmd *cobra.Command, args []string) error {
+		dryRun, _ := command.Flags().GetBool("dry-run")
+
 		if len(args) == 0 {
 			output, _ := command.Flags().GetString("output")
+
+			if dryRun {
+				fmt.Println(`Cannot use '--dry-run' flag without tool name.
+Get help using:
+arkade get --help`)
+				os.Exit(1)
+			}
 
 			if len(output) > 0 {
 				if get.TableFormat(output) == get.MarkdownStyle {
@@ -116,21 +127,25 @@ and provides a fast and easy alternative to a package manager.`,
 				operatingSystem,
 				version,
 				dlMode,
-				progress)
+				progress,
+				dryRun)
 			if err != nil {
 				return err
 			}
 
-			localToolsStore = append(localToolsStore, get.ToolLocal{Name: tool.Name, Path: outFilePath})
-			fmt.Printf("\nTool written to: %s\n\n", outFilePath)
+			if !dryRun {
+				localToolsStore = append(localToolsStore, get.ToolLocal{Name: tool.Name, Path: outFilePath})
+				fmt.Printf("\nTool written to: %s\n\n", outFilePath)
+			}
 		}
 
-		msg, err := get.PostInstallationMsg(dlMode, localToolsStore)
-		if err != nil {
-			return err
+		if !dryRun {
+			msg, err := get.PostInstallationMsg(dlMode, localToolsStore)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("%s", msg)
 		}
-
-		fmt.Printf("%s", msg)
 
 		return err
 	}
--- a/pkg/get/download.go
+++ b/pkg/get/download.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/alexellis/arkade/pkg/archive"
 	"github.com/alexellis/arkade/pkg/config"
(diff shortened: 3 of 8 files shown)


dirien:
LGTM!

alexellis:
The only issue I have with this, is that the output is the same as a regular download, including the extraction of files.

I expected `--dry-run` to print which URL/version would be downloaded and then exit.

The message saying how to run `dive` for instance, would be invalid if dive hadn't actually been downloaded.

pkg/get/download.go
```diff
@@ -33,7 +34,7 @@ func Download(tool *Tool, arch, operatingSystem, version string, downloadMode in
 
 	fmt.Printf("Downloading: %s\n", downloadURL)
```
alexellis, line 35: This is the point at which I would expect the code to exit in a dry-run.

yankeexe:
> The only issue I have with this, is that the output is the same as a regular download, including the extraction of files.
> 
> I expected `--dry-run` to print which URL/version would be downloaded and then exit.
> 
> The message saying how to run `dive` for instance, would be invalid if dive hadn't actually been downloaded.

I think it makes sense that `dry-run`  provides the user an end-to-end preview of what running that command looks like. That's what I thought of while creating the PR. What do you think? 

The user gets to see: 
1. download URL with version 
2. how it will get extracted
3. where it will get stored
4. what commands to run after the installation.
@alexellis

yankeexe:
@alexellis Update the PR, now the logs stop after showing the Download URL.

cmd/get.go
```diff
@@ -50,11 +51,21 @@ and provides a fast and easy alternative to a package manager.`,
...
 
 	command.RunE = func(cmd *cobra.Command, args []string) error {
+		dry_run, _ := command.Flags().GetBool("dry-run")
+
 		if len(args) == 0 {
 			output, _ := command.Flags().GetString("output")
 
+			if dry_run {
+				fmt.Println(`Cannot use '--dry-run' flag without tool name.
+Get help using:
```
rgee0, line 64:
```suggestion
Find help on the `get` command by using:
```
This is subtle and nitty but I think it will help newer users understand the nature of `--help`.   Further explanation would be from the opposite angle.... an unfamiliar user might misunderstand that the Arkade help command is more idiomatic `arkade 'get help'`, which could cause them to miss the rest of the help scope.

README.md
```diff
@@ -116,6 +116,9 @@ With automatic detection of: Windows / MacOS / Linux / Intel / ARM.
 arkade get APP
 arkade get APP --version VERSION
 arkade get APP@VERSION
+
+# Dry Run
+arkade get APP@VERSION --dry-run
```
rgee0, line 121: This is a wider observation, which may need broader consideration.  The terminology used here is inconsistent with that used within arkade itself.  Is this a conscious decision? For example, is there a distinction between when `app` is to be used versus when `tool` is used?  If the two terms are interchangeable then I would suggest choosing one, again for the benefit of unfamiliar users.

yankeexe: Yes, I second this, should be `tool` for a clearer distinction.

yankeexe:
@rgee0 I guess, it's best that we create new issues to address your comments and pick them up from there. :slightly_smiling_face:

README.md
alexellis: Yes "app" is a Kubernetes application.

Then we have tools / CLIs for downloads.

pkg/get/download.go
```diff
@@ -31,6 +32,24 @@ func Download(tool *Tool, arch, operatingSystem, version string, downloadMode in
...
 
+	// Check user provided version is valid on dry run
+	if len(version) > 1 && dry_run {
+		res, err := http.Head(downloadURL)
+		if err != nil {
+			return "", "", err
+		}
+
+		if res.StatusCode != http.StatusOK {
+			return "", "", fmt.Errorf("incorrect status for downloading tool: %d", res.StatusCode)
```
alexellis, line 43: I'm not sure that we want "incorrect status", this should indicate that the tool is not available or not found.

cmd/get.go
```diff
@@ -50,11 +51,21 @@ and provides a fast and easy alternative to a package manager.`,
 	command.Flags().StringP("output", "o", "", "Output format of the list of tools (table/markdown/list)")
 	command.Flags().Bool("stash", true, "When set to true, stash binary in HOME/.arkade/bin/, otherwise store in /tmp/")
 	command.Flags().StringP("version", "v", "", "Download a specific version")
+	command.Flags().Bool("dry-run", false, "When set to true, performs a dry run of binary download and install location")
 
 	command.RunE = func(cmd *cobra.Command, args []string) error {
+		dry_run, _ := command.Flags().GetBool("dry-run")
```
alexellis, line 57: Variables should be written `dryRun` in Go, this looks more like a Ruby / Python / PHP variable name.

pkg/get/download.go
```diff
@@ -21,7 +21,8 @@ const (
 	DownloadArkadeDir = iota
 )
 
-func Download(tool *Tool, arch, operatingSystem, version string, downloadMode int, displayProgress bool) (string, string, error) {
+func Download(tool *Tool, arch, operatingSystem, version string, downloadMode int, displayProgress bool, dry_runs ...bool) (string, string, error) {
+	dry_run := dry_runs[0]
```
alexellis, line 25: Why is there a slice of dry_runs, where you then only index the first item, and assume that there's always one item there? What if there were no items?

pkg/get/download.go
```diff
@@ -31,6 +32,24 @@ func Download(tool *Tool, arch, operatingSystem, version string, downloadMode in
 		return "", "", err
 	}
 
+	// Check user provided version is valid on dry run
+	if len(version) > 1 && dry_run {
```
alexellis, line 36: Same comment on the variable name.

pkg/get/download.go
```diff
@@ -31,6 +32,24 @@ func Download(tool *Tool, arch, operatingSystem, version string, downloadMode in
...
+
+		if res.StatusCode != http.StatusOK {
+			return "", "", fmt.Errorf("incorrect status for downloading tool: %d", res.StatusCode)
+		}
+
+	}
+
+	if dry_run {
+		fmt.Printf("Download URL: %s\n", downloadURL)
+		os.Exit(0)
```
alexellis, line 50: I would not expect to see an os.Exit within a package, if the Github diff is working right, we are looking at pkg/get/download.go

A return statement would make more sense.

pkg/get/download.go
```diff
@@ -31,6 +32,24 @@ func Download(tool *Tool, arch, operatingSystem, version string, downloadMode in
 		return "", "", err
 	}
 
+	// Check user provided version is valid on dry run
+	if len(version) > 1 && dry_run {
+		res, err := http.Head(downloadURL)
```
alexellis, line 37: Can you extract this code block into its own method with a descriptive name?

I'd also like to see you use http.NewRequest instead of the shortcut http.Head. See how we make HTTP requests elsewhere in the code for an example.

You'll then want `http. MethodHead`, https://pkg.go.dev/net/http#pkg-constants

alexellis (changes requested):
Some comments on the implementation and structure to be addressed before merge. Thank you for working on this feature.

yankeexe:
@alexellis ready for merge

Issue #527, opened 2021-10-02
[App] Add CockroachDB Helm Chart to apps


Shikachuu:
## Expected Behaviour
`arkade install cockroachdb` should install CockroachDB to the given k8s cluster.

## Current Behaviour
Currently not implemented.

## Are you a GitHub Sponsor (Yes/No?)

Check at https://github.com/sponsors/alexellis

- [ ] Yes
- [X] No

## Possible Solution
Add a new entry to the install command.

## Context

CockroachDB is a Postgres compatible database with multi-master and multi-cluster replication.
I want to deploy it to a k3d cluster installed by arkade, so it is an obvious choice to install it from arkade. :) 

## Your Environment

* What Kubernetes distribution are you using?

```
k3d, running k3s:
v1.22.2
```
* Operating System and version (e.g. Linux, Windows, MacOS):

```
Linux pop-os 5.13.0-7614-generic #14~1631647151~21.04~930e87c-Ubuntu SMP Fri Sep 17 00:24:58 UTC  x86_64 x86_64 x86_64 GNU/Linux
NAME="Pop!_OS"
VERSION="21.04"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 21.04"
VERSION_ID="21.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=hirsute
UBUNTU_CODENAME=hirsute
LOGO=distributor-logo-pop-os
```

* What arkade version is this?

```
0.8.2
```

alexellis:
Hi @Shikachuu I don't want you to do unnecessary work, so just a quick reminder how contributing works.

1) Suggest idea/enhancement
2) Wait for approval and feedback from a maintainer
3) Raise a PR

If I am remembering correctly, we have skipped 2) a few times.

Alex

alexellis:
You have been lucky in this instance, the app seems like a good fit, so I've left a review on the PR.

Shikachuu:
> Hi @Shikachuu I don't want you to do unnecessary work, so just a quick reminder how contributing works.
> 
>     1. Suggest idea/enhancement
> 
>     2. Wait for approval and feedback from a maintainer
> 
>     3. Raise a PR
> 
> 
[...]

PR #528, opened 2021-10-02
Add CockroachDB Helm Chart to the apps

Shikachuu:
Signed-off-by: Shikachuu <zcmate@gmail.com>

## Description
Added new CockroachDB helm chart to the installable apps.

## Motivation and Context

- [X] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md)) #527

- [ ] Yes
- [X] No

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Documentation

- [X] I have updated the list of tools in README.md if (required) with `./arkade get -o markdown`
- [X] I have updated the list of apps in README.md if (required) with `./arkade install --help`

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide


alexellis:
Hi @Shikachuu please can you run `git config user.name FULL NAME HERE` with your full name, so that we have that instead of your GitHub username in the sign-off? Thanks. Then run `git commit -s --amend` and force push.

alexellis:
You've missed a checklist item:

 I have tested this on arm, or have added code to prevent deployment

Can you confirm whether this app works on arm, and if not add the code you'll find in other apps to prevent its installation. Looking at the Docker Hub images is a good indicator and viewing the architectures available for the images. The docs for upstream projects also can be a good place to find out whether support exists.

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+		Use:          "cockroachdb",
+		Short:        "Install cockroachdb",
+		Long:         "Install cockroachdb",
+		Example:      ` arkade app install cockroachdb`,
+		SilenceUsage: true,
+	}
+	cockroachdb.Flags().String("namespace", "default", "Namespace for the app")
+	cockroachdb.Flags().Bool("persistence", false, "Use a 100Gi Persistent Volume to store data")
+
+	cockroachdb.Flags().Bool("single-node", false, "Run CockroachDB instances in standalone mode with replication disabled, so the StatefulSet does NOT FORM A CLUSTER")
```
alexellis, line 26: Would `single-node=true` be more intuitive as a flag named `cluster` true/false?

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+	cockroachdb.Flags().Int64("replicas", 1, "Satetfulset replica count")
```
alexellis, line 27: `Satetfulset` is a typo.

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+
+	cockroachdb.Flags().Bool("enable-tls", false, "Whether to run securely using TLS certificates")
```
alexellis, line 29: Shorten to `tls`

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+		enableTls, _ := command.Flags().GetBool("enable-tls")
+		replicas, _ := command.Flags().GetInt64("replicas")
+
+		overrides := map[string]string{}
+
+		if singleNode {
+			overrides["conf.single-node"] = "true"
+		}
+		if enableTls {
+			overrides["tls.enabled"] = "false"
```
alexellis, line 51: Does this not require additional settings like Ingress being turned on, or does it work through a TCP connection?

Is it a self-signed cert?

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		println(cockroachdbInstallMsg)
+		return nil
+	}
+
+	return cockroachdb
+}
+
+const CockroachdbInfoMsg = `# Get started at: https://www.cockroachlabs.com/docs/stable
```
alexellis, line 83: Is there a way to get logs?

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+
+# If you used tls check that the secrets were created on the cluster
+
+kubectl get secrets
```
alexellis, line 87: What is the name of the secret, is it stable and can we add it here?

cmd/apps/cockroachdb_app.go
```diff
@@ -0,0 +1,100 @@
...
+
+# If you used persistence confirm that the volumes are created
+
+kubectl get pv
+
+# Visit the CockroachDB dashboard:
+
+kubectl port-forward service/cockroachdb 8080`
```
alexellis, line 95: Can you write it as `8080:8080`, so that people can change the initial port more easily if there is a conflict?

alexellis (changes requested):
Looks promising. I have a few questions about user-experience.

We also need to confirm whether arm is supported and block it if not.

cmd/apps/cockroachdb_app.go
Shikachuu: This is TLS for the database connection itself, not for the dashboard.
It is self-signed, but has options in the values to provide one, so in this case you can provide it with the `--set` flag.

cmd/apps/cockroachdb_app.go
Shikachuu: I actually inherited the flag `single-node` from the upstream project.
The `cockroach` binary uses the same flag with the `values.yml` aswell.

Based on this facts I think it is better to keep it this way.

cmd/apps/cockroachdb_app.go
Shikachuu: Those are the self signed tls certs, that cockroach gives you if you use tls auth for accessing the database itself.

cmd/apps/cockroachdb_app.go
Shikachuu: Sure thing.

cmd/apps/cockroachdb_app.go
Shikachuu: This is make sense, changed in the latest commit.

cmd/apps/cockroachdb_app.go
Shikachuu: Nice catch, fixed.

Shikachuu: Nice catch, thanks, fixed.

Shikachuu:
> You've missed a checklist item:
> 
> I have tested this on arm, or have added code to prevent deployment
> 
> Can you confirm whether this app works on arm, and if not add the code you'll find in other apps to prevent its installation. Looking at the Docker Hub images is a good indicator and viewing the architectures available for the images. The docs for upstream projects also can be a good place to find out whether support exists.

Thanks for the notice, 
currently there is no arm build for this, I will add something that will prevent the install process asap.

cmd/apps/cockroachdb_app.go
Shikachuu: Are we talking about install logs or application logs?

cmd/apps/cockroachdb_app.go
alexellis: How does the user download the TLS secret and put it into their trust bundle? Do we need a command for that?

cmd/apps/cockroachdb_app.go
alexellis: This means the application logs of course. The logs from helm are always printed on the console.

cmd/apps/cockroachdb_app.go
alexellis: Same question: what is the name of the secret, can you add the specific name here please, `kubectl get secrets` will not necessarily make it obvious.

Look at how we do this in the openfaas app for instance.

alexellis:
I've replied to some of the unresolved comments. The if-statement for arm could probably be taken from the Istio app.

cmd/apps/cockroachdb_app.go
Shikachuu: Logs are disabled in the helm chart by default. So until you enable it I don't think you can.

cmd/apps/cockroachdb_app.go
Shikachuu: Sorry I missunderstood the question, sure I will provide a more detailed command here, thanks for the notice! :)

cmd/apps/cockroachdb_app.go
Shikachuu: If TLS is enabled, you should download the CA cert and you should put it into your `.postgresql` folder in your application user's home directory, and you should use the client certificate in your application to connect to the DB, so in case of golang your should put it into your `pgx` connection.

From my perspective, this message is all about how to install and manage the installed application from the devops side.
I don't think this is in the scope of installing or managing the cluster, more like how do you use it in code.
Am I wrong in this case?

Shikachuu:
> I've replied to some of the unresolved comments. The if-statement for arm could probably be taken from the Istio app.

Thanks for the tip, I had to figure it out what it exactly does since I don't want to just plain copy + paste without knowing what I am doing. :smile: 
But I did it and it is working flawlessly.

cmd/apps/cockroachdb_app.go
alexellis: `kubectl get secrets` doesn't help the user do the right thing, it'd almost be better to omit it.

If we can get the specific secret name and decode it for them it'd be more intuitive and a better experience for a first-time user.

alexellis:
Once you've addressed https://github.com/alexellis/arkade/pull/528#discussion_r722002717

Can you then rebase and squash down these commits please?

https://www.youtube.com/watch?v=8j0H6urZ-bU

Issue #571, opened 2021-10-28
Feature request: Add Kuma helm chart + kumactl cli app


Shikachuu:
Add `kumactl` to `arkade get` and add Kuma itself to `arkade install`.

## Expected Behaviour
`arkade install kuma` should install the Kuma helm chart and `arkade get kumactl` should install kumactl.

## Current Behaviour
Neither the helm chart nor kumactl implemented.

## Are you a GitHub Sponsor (Yes/No?)

Check at https://github.com/sponsors/alexellis

- [ ] Yes
- [X] No

## Possible Solution
Add kumactl to cli and the helm chart to arkade.

## Steps to Reproduce (for bugs)
This isn't a bug.

## Your Environment

* What Kubernetes distribution are you using?

Kubectl and OS version shouldn't matter since this is not implemented yet.

* What arkade version is this?

```
Version: 0.8.8
Git Commit: 10f4f5e0b2f5012b85a7a2b4b45a0aecdb4d8ea6
```

alexellis:
Approved 👍

Shikachuu:
/edit title: Feature request: Add Kuma helm chart + kumactl cli app

PR #572, opened 2021-10-30
Add kumactl to arkade get

Shikachuu:
## Description
Related to #571, it adds the Kuma cli tool to `arkade get`

## Motivation and Context

- [X] I have raised an issue to propose this change, which has been given a label of `design/approved` by a maintainer ([required](https://github.com/alexellis/arkade/blob/master/CONTRIBUTING.md)) #571

## How Has This Been Tested?
After build I ran `arkade get kumactl` and the install worked on the following systems:
- Debian
- CentOS
- Ubuntu
- RedHat

**It hasn't been tested on darwin, since currently I do not own a Mac.**

- [ ] Yes
- [X] No

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Documentation

- [X] I have updated the list of tools in README.md if (required) with `./arkade get -o markdown`

## Checklist:

- [X] I've read the [CONTRIBUTION](https://github.com/alexellis/arkade/blob/master/CONTRIBUTING.md) guide

### DIFF
--- a/pkg/get/tools.go
+++ b/pkg/get/tools.go
@@ -2087,5 +2087,29 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Name}}
 			{{$name}}
 	`})
 
+	tools = append(tools,
+		Tool{
+			Owner:       "kumahq",
+			Repo:        "kuma",
+			Name:        "kumactl",
+			Version:     "1.4.1",
+			Description: "kumactl is a CLI to interact with Kuma and its data",
+			URLTemplate: `
+			{{$osStr := ""}}
+			{{$archStr := ""}}
+			{{- if HasPrefix .OS "linux" -}}
+			{{$osStr = "ubuntu"}}
+			{{- else if eq .OS "darwin" -}}
+			{{$osStr = "darwin"}}
+			{{- end -}}
+
+			{{- if eq .Arch "x86_64" -}}
+			{{$archStr = "amd64"}}
+			{{- else -}}
+			{{$archStr = .Arch}}
+			{{- end -}}
+			https://download.konghq.com/mesh-alpine/{{.Repo}}-{{.Version}}-{{$osStr}}-{{$archStr}}.tar.gz`,
+			BinaryTemplate: `{{.Name}}`,
+		})
 	return tools
 }
--- a/pkg/get/get_test.go
+++ b/pkg/get/get_test.go
@@ -3418,3 +3418,37 @@ func Test_DownloadDagger(t *testing.T) {
 		}
 	}
 }
+
+func Test_DownloadKumactl(t *testing.T) {
+	tools := MakeTools()
+	name := "kumactl"
+
+	tool := getTool(name, tools)
+
+	tests := []test{
+		{
+			os:      "darwin",
+			arch:    arch64bit,
+			version: "1.4.1",
+			url:     "https://download.konghq.com/mesh-alpine/kuma-1.4.1-darwin-amd64.tar.gz",
+		},
+		{
+			os:      "linux",
+			arch:    arch64bit,
+			version: "1.4.1",
+			url:     "https://download.konghq.com/mesh-alpine/kuma-1.4.1-ubuntu-amd64.tar.gz",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.os+" "+tc.arch+" "+tc.version, func(r *testing.T) {
+			got, err := tool.GetURL(tc.os, tc.arch, tc.version, false)
+			if err != nil {
+				t.Fatal(err)
+			}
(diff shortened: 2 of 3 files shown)


pkg/get/tools.go
```diff
@@ -1823,5 +1823,23 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Name}}
...
 		})
 
+	tools = append(tools,
+		Tool{
+			Owner:       "kumahq",
+			Repo:        "kuma",
+			Name:        "kumactl",
+			Version:     "1.3.1",
+			Description: "kumactl is a CLI to interact with Kuma and its data",
+			URLTemplate: `
```
alexellis, line 1833: Looks good. We will need a unit test adding though. I don't mind if you do that as commit 2 or reset into one commit.

alexellis (changes requested):
Please see my message.

You can test these by running:

```
# Override each variation, including one which is unavailable:
arkade get --os ... --arch ...

# Print the architecture:
file ./kumactl
```

pkg/get/tools.go
Shikachuu: I added some test for darwin x64 and linux x64 (only these arch + os combos are supported).

pkg/get/tools.go
```diff
@@ -1823,5 +1823,23 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Name}}
...
+			{{$osStr := ""}}
+			{{- if HasPrefix .OS "linux" -}}
+			{{$osStr = "ubuntu"}}
+			{{- else if eq .OS "darwin" -}}
+			{{$osStr = "darwin"}}
+			{{- end -}}
+			https://download.konghq.com/mesh-alpine/{{.Repo}}-{{.Version}}-{{$osStr}}-amd64.tar.gz`,
```
alexellis, line 1840: So, if you use armv7l as the `--arch` override, you will get the wrong binary.

We need to template the `amd64` portion of this string. Why? So that's empty unless it's set otherwise.

For invalid architectures and operating systems we should cause a 404 in the URL.

Also, do they really not have a Windows binary for their CLI?

alexellis:
The video I have on rebasing commits into one might be useful. Derek shared a link for it, you can squash the no dco error at the same time.

pkg/get/tools.go
Shikachuu: They do not have a windows binary. :disappointed: 
I'll fix the arch problems asap, thanks for the note! :smile:

Shikachuu:
> The video I have on rebasing commits into one might be useful. Derek shared a link for it, you can squash the no dco error at the same time.

It was a little more complex since I auto-merged master before, so had to do some tinkering but it came up just fine. :smile:

alexellis:
Please rebase @Shikachuu

alexellis:
We need a quick rebase on this one @Shikachuu - please also see if there's anything else you want to do for consistency, or if it's good to go.

alexellis (approved):
Approved

alexellis:
@Shikachuu please can you rebase due to other merged PRs conflicting with these changes?

Shikachuu:
Done! :heavy_check_mark: 
But the e2e url check failed...yet the URL is alive and well.

alexellis:
>But the e2e url check failed...yet the URL is alive and well.

Can you work out why?

alexellis:
Run the command locally, and you'll see that it's forbidden:

```
curl -i --head -L https://download.konghq.com/mesh-alpine/kuma-1.4.1-ubuntu-amd64.tar.gz
```

This is probably because Kuma is using an S3 bucket, perhaps it disallows HEAD methods?

Shikachuu:
> Run the command locally, and you'll see that it's forbidden:
> 
> ```
> curl -i --head -L https://download.konghq.com/mesh-alpine/kuma-1.4.1-ubuntu-amd64.tar.gz
> ```
> 
> This is probably because Kuma is using an S3 bucket, perhaps it disallows HEAD methods?
> 
> https://stackoverflow.com/questions/27469801/403-forbidden-error-on-s3-rest-api-head-request

As we talked about this, the S3 bucket disallows HEAD requests, so I added a skip mechanism to the e2e tests, and added `kumactl` to it. [Also I created an issue in the kuma main repo as well.](https://github.com/kumahq/kuma/issues/4019)

dannysauer:
I'm linking [my comment with workarounds](https://github.com/kumahq/kuma/issues/4019#issuecomment-1075399577) in a new comment here since I forgot to tag this issue for cross-thread awareness. :)  If there's any feedback on feasibility of those workarounds which might help other people, I guess the Kong thread might be the more discoverable place to provide such feedback.  Thanks!

alexellis (approved):
Approved

Issue #654, opened 2022-04-01
Feature: system packages for Linux servers, CI and workstations


alexellis:
## Feature

Feature: system packages for Linux servers and workstations

There are so many times that I log into a server and install server software using a combination of curl and tar.

This provides the fastest way to install software, ensures the latest versions are available, something that package managers struggle to keep up with.

Generally, during scripting and automation, you'll find this kind of pattern. The popularity of `arkade get` has shown that people are comfortable installing software in this way, and it makes CI - trivial - https://github.com/hopefulramen/k3sup-tester/blob/master/.github/workflows/build.yml 

For example:

```bash
arkade system install go@v1.18.1
arkade system install node@v1.16
arkade system install firecracker@v1.0.0
arkade system install containerd@v1.16.1
arkade system install cni-plugins
```

This is present in openfaas/faasd, and on every Linux workstation I use, I tend to need these tools.

When a tool is already present, installation will be skipped, unless `--force` is passed.

The version will be detected the same way as `arkade get` without using the GitHub API, but using a HTTP HEAD request to the releases page, where no version is given that is.

A Go template will be used to form the URL, which will have a unit test for ARM32, ARM64 and AMD64.

Sometimes there are additional steps to run like systemd enable, or making a directory in a known location.

This would be a new command aimed at ARM32, ARM64 and AMD/Intel workstations and servers, and would save a lot of time for CI and setting up workstations.

I'd suggest we create a basic fluid SDK when adding the first app for Golang, for instance:

```bash
export ARCH=$(uname -p)

export GOPATH=$HOME/go/
export PATH=$PATH:/usr/local/go/bin/

install_go() {
    DL_ARCH=$ARCH
    if [ "$ARCH" == "x86_64" ]; then
        DL_ARCH=amd64
    else 
        if [ "$ARCH" == "aarch64" ]; then
            DL_ARCH=arm64
        fi
    fi

    if ! [ -e /usr/local/go/bin/go ]; then

        curl -SLf https://go.dev/dl/go$GOVER.linux-$DL_ARCH.tar.gz > /tmp/go.tgz
        sudo rm -rf /usr/local/go/
        sudo mkdir -p /usr/local/go/
        sudo tar -xvf /tmp/go.tgz -C /usr/local/go/ --strip-components=1

        go version
        echo "export GOPATH=\$HOME/go/" | tee -a $HOME/.bashrc
[...]

PR #663, opened 2022-04-05
Add system packages and an app for Go

alexellis:
## Description

Add system packages and an app for Go

```bash
arkade system install go --version 1.18
```

## Motivation and Context

#654 

## How Has This Been Tested?

```bash
$ go build && ./arkade system install go --path /tmp/go-118
Installing Go to /tmp/go-118
Installing version: go1.18 for: amd64
Downloading from: https://go.dev/dl/go1.18.linux-amd64.tar.gz
135.14 MiB / 135.14 MiB [---------------------------------------------------------------------------] 100.00%
Downloaded to: /tmp/go1.18.linux-amd64.tar.gz
Unpacking Go to: /tmp/go-118/go
2022/04/05 09:29:41 extracted tarball into /tmp/go-118: 11753 files, 1378 dirs (2.254868602s)

export PATH=$PATH:/tmp/go-118/go/bin:$HOME/go/bin
export GOPATH=$HOME/go/
$ /tmp/go-118/go/bin/go version
go version go1.18 linux/amd64
```

## Types of changes

## Documentation

I've updated the README

## Checklist:

### DIFF
--- a/cmd/system/go.go
+++ b/cmd/system/go.go
@@ -1,0 +1,130 @@
+// Copyright (c) arkade author(s) 2022. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package system
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/alexellis/arkade/pkg/archive"
+	"github.com/alexellis/arkade/pkg/env"
+	"github.com/alexellis/arkade/pkg/get"
+	"github.com/spf13/cobra"
+)
+
+func MakeInstallGo() *cobra.Command {
+
+	command := &cobra.Command{
+		Use:   "go",
+		Short: "Install Go",
+		Long:  `Install Go programming language and SDK.`,
+		Example: `  arkade system install go
+  arkade system install go --version v1.18.1`,
+		SilenceUsage: true,
+	}
+
+	command.Flags().StringP("version", "v", "", "The version for Go, or leave blank for pinned version")
+	command.Flags().String("path", "/usr/local/", "Installation path, where a go subfolder will be created")
+	command.Flags().Bool("progress", true, "Show download progress")
+
+	command.PreRunE = func(cmd *cobra.Command, args []string) error {
+
+		return nil
+	}
+
+	command.RunE = func(cmd *cobra.Command, args []string) error {
+		installPath, _ := cmd.Flags().GetString("path")
+		version, _ := cmd.Flags().GetString("version")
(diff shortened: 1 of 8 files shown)


cmd/system/system.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/spf13/cobra"
+)
+
+func MakeSystem() *cobra.Command {
+
+	command := &cobra.Command{
+		Use:     "system",
+		Short:   "System apps",
+		Long:    `Apps for systems.`,
+		Aliases: []string{"v"},
```
Shikachuu, line 17: Why `v`?

pkg/archive/untar_nested.go
```diff
@@ -0,0 +1,129 @@
...
+	"path/filepath"
+	"time"
+)
+
+// Untar reads the gzip-compressed tar file from r and writes it into dir.
+func UntarNested(r io.Reader, dir string) error {
+	return untarNested(r, dir)
+}
+
+func untarNested(r io.Reader, dir string) (err error) {
```
Shikachuu, line 19: Should we move this functions under `pkg/archive/untar.go`?
I think it makes much more sense.

cmd/system/go.go
```diff
@@ -0,0 +1,130 @@
...
+
+	command.PreRunE = func(cmd *cobra.Command, args []string) error {
+
+		return nil
+	}
+
+	command.RunE = func(cmd *cobra.Command, args []string) error {
+		installPath, _ := cmd.Flags().GetString("path")
+		version, _ := cmd.Flags().GetString("version")
+		fmt.Printf("Installing Go to %s\n", installPath)
```
Shikachuu, line 43: Should we introduce `--quiet` flag as we did earlier for binary downloads?

cmd/system/system.go
alexellis: I copied from the venafi app so we can change that to `s`

pkg/archive/untar_nested.go
alexellis: No I don't see any reason to have one file for this.

cmd/system/go.go
alexellis: I'm going to keep things verbose for the preview stages, to make it simpler / less work for everyone.

Then we can come back and fine tune.

alexellis:
@Shikachuu do you want to have a go at the containerd recipe and prometheus next?

@richardcase volunteered to take the firecracker CLI

I didn't use a Go template in this Go app, but if we need it we can bring that back in for the additional apps we're working with.

Shikachuu:
@alexellis sure I'll play around with containerd!

PR #667, opened 2022-04-06
feat: add firecracker install

richardcase:
This adds the ability to download and install Firecracker.

## Description
This adds the following to enable user to install Firecracker as a system package:

```bash
arkade system install firecracker
```

## Motivation and Context

Relates to #654 

## How Has This Been Tested?
Local testing in my development env but not on arm:

```bash
➜ go build && ./arkade system install firecracker --path ./out
Installing Firecracker to ./out
Installing version: v1.0.0 for: x86_64
Downloading from: https://github.com/firecracker-microvm/firecracker/releases/download/v1.0.0/firecracker-v1.0.0-x86_64.tgz
1.59 MiB / 1.59 MiB [------------------------------------------------------------------------------------------------------------------------------------] 100.00%
Downloaded to: /tmp/firecracker-v1.0.0-x86_64.tgz
Unpacking Firecracker to: /tmp/firecracker3308092613
Copying Firecracker binaries to: ./out

➜ out/firecracker --version
Firecracker v1.0.0

Supported snapshot data format versions: v0.23.0, v0.24.0, v0.25.0, v1.0.0
```

- [ ] Yes
- [x] No

## Types of changes

## Checklist:

### DIFF
--- a/cmd/system/firecracker.go
+++ b/cmd/system/firecracker.go
@@ -1,0 +1,119 @@
+// Copyright (c) arkade author(s) 2022. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package system
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/alexellis/arkade/pkg/archive"
+	"github.com/alexellis/arkade/pkg/env"
+	"github.com/alexellis/arkade/pkg/get"
+	"github.com/spf13/cobra"
+)
+
+const (
+	githubDownloadTemplate = "https://github.com/%s/%s/releases/download/%s/%s"
+	githubLatest           = "latest"
+
+	firecrackerOwner = "firecracker-microvm"
+	firecrackerRepo  = "firecracker"
+
+	repoFlagName     = "repo"
+	versionFlagName  = "version"
+	pathFlagName     = "path"
+	progressFlagName = "progress"
+)
+
+func MakeInstallFirecracker() *cobra.Command {
+
+	command := &cobra.Command{
+		Use:   "firecracker",
+		Short: "Install Firecracker",
+		Long:  `Install Firecracker and its Jailer.`,
+		Example: `  arkade system install firecracker
+  arkade system install firecracker --version v1.0.0`,
+		SilenceUsage: true,
+	}
+
+	command.Flags().StringP(versionFlagName, "v", githubLatest, "The version for Firecracker to install")
+	command.Flags().StringP(pathFlagName, "p", "/usr/local/bin", "Installation path, where a go subfolder will be created")
+	command.Flags().Bool(progressFlagName, true, "Show download progress")
+
+	command.RunE = func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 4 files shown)


alexellis:
Thanks for the PR.

I'd like to see firecracker installed as `firecracker`, with a default path of `/usr/local/bin`

This is how I install it presently:

```bash
install_firecracker() {
    if ! [ -e /usr/local/bin/firecracker ] ; then 
        curl -L -S -f -o /tmp/firecracker.tgz https://github.com/firecracker-microvm/firecracker/releases/download/v$FIRECRACKER_VER/firecracker-v$FIRECRACKER_VER-$ARCH.tgz
        if [ $? -ne 0 ]; then
            echo "Failed to download firecracker"
            exit 1
        fi
        sudo tar -xvf /tmp/firecracker.tgz -C /usr/local/bin/ --strip-components=1
        sudo mv /usr/local/bin/{firecracker-v$FIRECRACKER_VER-$ARCH,firecracker}
        sudo mv /usr/local/bin/{jailer-v$FIRECRACKER_VER-$ARCH,jailer}
        sudo mv /usr/local/bin/{seccompiler-bin-v$FIRECRACKER_VER-$ARCH,seccompiler-bin}
        sudo mv /usr/local/bin/{seccomp-filter-v$FIRECRACKER_VER-$ARCH.json,seccomp-filter.json}
    fi

}
```

Alex

cmd/system/firecracker.go
```diff
@@ -0,0 +1,145 @@
...
+
+		return nil
+	}
+
+	return command
+}
+
+func getLatestGitHubRelease(repo string) (string, error) {
+
+	latestUrl := fmt.Sprintf(gitHubLatestTemplate, repo)
```
alexellis, line 115: This mechanism will be rate limited, instead we use the HTTP API and already have methods to do this that you can re-use.

Please remove and find the equivalent package from pkg/ instead.

Thanks

alexellis (changes requested):
Looking forward to getting this in. I've left a couple of notes.

cmd/system/firecracker.go
richardcase: Changed.

cmd/system/firecracker.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		fmt.Printf("Downloaded to: %s\n", outPath)
+
+		f, err := os.OpenFile(outPath, os.O_RDONLY, 0644)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		tempUnpackPath, err := os.MkdirTemp(os.TempDir(), "firecracker*")
```
alexellis, line 95: Would we benefit from a os.Remove/Dir for tempUnpackPath ?

richardcase: Yes i think you are right.

alexellis: Perhaps you could do that in the next app you contribute? @Shikachuu is covering containerd but the rest are fair game.

Node should be fairly similar to add?

https://nodejs.org/en/download/

PR #932, opened 2023-06-14
Hoist test runner in get_test.go

joebowbeer:
Removes 1000 duplicated lines from `get_test.go`

## Description
Extract the GetURL test runner loop into a new function.

## Motivation and Context
Eliminates duplication, and makes it easier to add enhancements to all tests.

## How Has This Been Tested?
make e2e
PASS

## Documentation

## Checklist:
- [ ] I've read the [CONTRIBUTION](https://github.com/alexellis/arkade/blob/master/CONTRIBUTING.md) guide
- [ ] I have signed-off my commits with `git commit -s`

### DIFF
--- a/pkg/get/get_test.go
+++ b/pkg/get/get_test.go
@@ -35,6 +35,20 @@ func getTool(name string, tools []Tool) *Tool {
 	return tool
 }
 
+func runGetToolTemplateTests(t *testing.T, tool *Tool, tests []test) {
+	for _, tc := range tests {
+		t.Run(tc.os+" "+tc.arch+" "+tc.version, func(t *testing.T) {
+			got, err := tool.GetURL(tc.os, tc.arch, tc.version, false)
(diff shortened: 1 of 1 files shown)


joebowbeer:
> I'm not sure that I like this and the general direction you end up with, but get why you wanted to do it.
> 
> Thanks for suggesting it and showing what it'd look like.

It's just a default implementation, but that's fine.

Any objection to fixing inletsctl?

alexellis:
> It's just a default implementation, but that's fine.

Can you explain what you mean by "it's a default implementation"?

We let test helpers come into another project called openfaas/certifier and it started with only one, then got worse and worse until now it's hard to know what's even happening during a test.

Let's keep the duplication, knowing we can always refactor with a search and replace to this pattern in the future.

Thanks again for the suggestion

> Any objection to fixing inletsctl?

Of course not.

Issue #977, opened 2023-09-29
Faliures when running `arkade update` on windows


Johannestegner:
## Expected Behaviour

Clean update without any issues when running `arkade update` on windows.

## Current Behaviour

There are a couple of issues with the windows build:

1. The `replaceExec` function fails due to not being able to replace the file (it's in use), I have a potential solution to this.
2. The binary downloaded does not seem to be the correct one:

```
λ .\bin\arkade.exe update                                                                                                                                                                                                                                                                                                                                                                                                 
Downloading: https://github.com/alexellis/arkade/releases/download/0.10.9/arkade
9.41 MiB / 9.41 MiB [-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
Checksum verified..OK.
Replaced: D:\Repositories\arkade testing\bin\arkade.exe..OK.

 🚀 Speed up GitHub Actions/GitLab CI + reduce costs: https://actuated.dev

λ .\bin\arkade.exe version                                                                                                                                                                                                                                                                                                                                                                                                
This version of D:\Repositories\arkade testing\bin\arkade.exe is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
```

## Possible Solution

The file system issue can be fixed by not "defering" the close calls on the stream. Basically just moving the closes down and not defer, as the file have to be closed to be replaced.

```
	df, err := os.OpenFile(newExec, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0755)
	if err != nil {
		return err
	}

	if _, err := io.Copy(df, sf); err != nil {
		return err
	}
	_ = sf.Close()
	_ = df.Close()

	// Replace the current executable file with the new executable file
	if err := os.Rename(newExec, currentExec); err != nil {
		return err
	}

	return nil
```

_I can create a PR for this if wanted :)_

The other issue seems to be that it downloads the wrong file, probably a linux x86_64 build:

```
λ file bin\arkade.exe
bin\arkade.exe: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=HHrNyUA19A6y4A1W8PS7/nh6SLLgGRSxvvgqC7mg1/oSWVZmFk0jKeLKR3wcOy/vx_B4XmJqPCuFc8P7u1I, stripped
```

vs:

```
λ file bin\arkade.exe.old                             
[...]

PR #988, opened 2023-10-10
Updated GetClientArch to use runtime package instead of uname checks.

Johannestegner:
## Description

This pull request replaces the current code in the `GetClientArch` function with one depending on the `runtime.GOARCH` and `runtime.GOOS` constants instead of using `uname` to check arch and os.

## Motivation and Context

The primary motivation for this PR is the discussion in #977, and the fact that using goos/goarch will allow for the values resolving fine on windows without the use of the `git bash` tool.

## How Has This Been Tested?

This has been tested by building (`make dist`) the binaries with `dev` as version, then transferring the files to the following platforms: `darwin-arm64`, `darwin-amd64`, `linux-amd64`, `linux-armv7l`, `linux-arm64` and then run `arkade update` and make sure that the correct binary is downloaded.

## Types of changes

- [X] Refactoring

## Checklist:
- [X] I've read the [CONTRIBUTION](https://github.com/alexellis/arkade/blob/master/CONTRIBUTING.md) guide
- [X] I have tested this on arm, or have added code to prevent deployment


pkg/env/env.go
```diff
@@ -4,39 +4,26 @@
...
 
-	archResult := strings.TrimSpace(res.Stdout)
-
-	taskOS := execute.ExecTask{Command: "uname",
-		Args:        []string{"-s"},
-		StreamStdio: false}
-	resOS, errOS := taskOS.Execute(context.Background())
-	if errOS != nil {
-		log.Println(errOS)
+	if arch == "arm" {
```
Johannestegner, line 18: I'm not super happy with the `if`'s here. I kinda wanted to do some type of `map` to map it better and make it easier to extend in the future, but decided to start as simple as possible and improve upon request. So let me know if it is too ugly!

pkg/env/env.go
```diff
@@ -4,39 +4,26 @@
...
+	"runtime"
 	"strings"
-
-	execute "github.com/alexellis/go-execute/v2"
 )
 
 // GetClientArch returns a pair of arch and os
 func GetClientArch() (arch string, os string) {
-	task := execute.ExecTask{
-		Command:     "uname",
```
alexellis, line 19: I'm sorry but this is not what I was asking for.

There should be a separate version of GetClientArch() which is combined only for Windows which looks at runtime.GOOS and returns the correct value. It could even be hard-coded given that it will only build in for Windows.

Then leave the existing code as it is for "unix"

Here's an example of how it works in Go: https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

Johannestegner: Ahh, I must have misunderstood, will fix asap! :)

Johannestegner:
PR updated and `Type` changed to Breaking (see "remarks" in description).

Further, I included both `go:build` and `+build` directives, I'm not 100% sure this is needed, but wanted to make sure I don't break the build (let me know if the old `+build` directive should be removed).

Johannestegner:
The `GOARCH` variable in the runtime package will return `arm64` for windows, do you prefer this to adhere to the `uname` result (x86_64) instead? Currently, I don't think there is a huge difference, because I don't think any checks are done on the architecture for windows in the application, but for future proofing it, it might be a good idea to stick to one standard.

alexellis:
> Although this PR allows for usage of other terminals than git-bash/ming in windows, this will make the ming target break, this due to the fact that earlier, the uname command determined if it was windows with ming while any terminal on windows now will return windows.

> This will most likely require the ming check to be removed and replaced with a windows check (should this be included in this pull request)?

What if MING was returned from the function instead of "windows"?

Otherwise, it'd require a lot of search and replace and testing to make sure we'd covered all the various templates and their unit tests.

Johannestegner:
Replaced return value with hardcoded `ming` string, tested and it works fine both in windows terminal as well as git-bash :)

```
λ .\arkade.exe get yq
Downloading: yq
2023/10/24 09:06:06 Looking up version for yq
2023/10/24 09:06:07 Found: v4.35.2
Downloading: https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_windows_amd64.exe
```

pkg/env/env.go
yarons: Does is cover all `arm` variants?

Johannestegner: This code is not part of the latest commit :)

PR #992, opened 2023-10-15
feat: allow to set default ingressclass

### LINKED ISSUE #983: Add easy option to "arkade install ingress-nginx" to set default IngressClass
## Expected Behaviour
Add an easy option to `arkade install ingress-nginx` to set it as the default IngressClass.

More info about the use of default ingress class:
https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class

## Current Behaviour
When using arkade to install ingress-nginx, the added IngressClass is not set as the default ingress class.

Next, when creating a new Ingress defintion, if ingressClassName (or the deprecated annotation 'kubernetes.io/ingress.class: nginx') is omitted , the ingress is added without any error or warning, but it will not work.
The ingress controller pod logs something like:  error="ingress does not contain a valid IngressClass",

To set the IngressClass as the default when using `arkade install ingress-nginx`, a user would need to run the following command:
`arkade install ingress-nginx -n ingress-nginx --set controller.ingressClassResource.default=true`

## Possible Solutions
1)
Add an easy to use flag for `arkade install ingress-nginx`, similar to --host-mode.
So something like: `arkade install ingress-nginx --set-as-default`

Add flag to the help text:
```
Examples:
  arkade install ingress-nginx --namespace default --set-as-default

```
2)
Another solution would be to just add this as an example for the --set flags in the help text for `install ingress-nginx -h`:
```
      --set stringArray    Use custom flags or override existing flags
                           (example: --set image=org/repo:tag)
                           (example to set as default Ingress Class: --set controller.ingressClassResource.default=true)
```
But this way, this important setting, is a bit hidden.

## Your Environment

* What Kubernetes distribution are you using?

```
Client Version: v1.27.6
Kustomize Version: v5.0.1
Server Version: v1.27.6+k3s1
```

* Operating System and version (e.g. Linux, Windows, MacOS):

```
Debian Bookworm
```

```
Version: 0.10.13
```

Jasstkn:
## Description

allows to set default ingress class. useful in case of multiple nginx ingress controllers.
fixes #983 

## Motivation and Context

## How Has This Been Tested?

```
$ ./arkade install nginx-ingress
...

$ k describe ingressclass nginx
Name:         nginx
Labels:       app.kubernetes.io/component=controller
              app.kubernetes.io/instance=ingress-nginx
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=ingress-nginx
              app.kubernetes.io/part-of=ingress-nginx
              app.kubernetes.io/version=1.9.3
              helm.sh/chart=ingress-nginx-4.8.2
Annotations:  ingressclass.kubernetes.io/is-default-class: true
              meta.helm.sh/release-name: ingress-nginx
              meta.helm.sh/release-namespace: default
Controller:   k8s.io/ingress-nginx
Events:       <none>
```

## Types of changes

## Documentation

## Checklist:

### DIFF
--- a/cmd/apps/nginx_app.go
+++ b/cmd/apps/nginx_app.go
@@ -28,6 +28,7 @@ flag and the ingress-nginx docs for more info`,
 	nginx.Flags().StringP("namespace", "n", "default", "The namespace used for installation")
 	nginx.Flags().Bool("update-repo", true, "Update the helm repo")
 	nginx.Flags().Bool("host-mode", false, "If we should install ingress-nginx in host mode.")
+	nginx.Flags().Bool("default-ingress", true, "Is this the default ingressClass for the cluster?")
 	nginx.Flags().StringArray("set", []string{}, "Use custom flags or override existing flags \n(example --set image=org/repo:tag)")
 
 	nginx.RunE = func(command *cobra.Command, args []string) error {
@@ -55,6 +56,14 @@ flag and the ingress-nginx docs for more info`,
 			overrides["controller.service.type"] = "NodePort"
 			overrides["dnsPolicy"] = "ClusterFirstWithHostNet"
 			overrides["controller.kind"] = "DaemonSet"
+		}
+
+		defaultIngress, flagErr := command.Flags().GetBool("default-ingress")
+		if flagErr != nil {
+			return flagErr
+		}
+		if defaultIngress {
+			overrides["controller.ingressClassResource.default"] = "true"
 		}
 
 		customFlags, _ := command.Flags().GetStringArray("set")


cmd/apps/nginx_app.go
```diff
@@ -28,6 +28,7 @@ flag and the ingress-nginx docs for more info`,
 	nginx.Flags().StringP("namespace", "n", "default", "The namespace used for installation")
 	nginx.Flags().Bool("update-repo", true, "Update the helm repo")
 	nginx.Flags().Bool("host-mode", false, "If we should install ingress-nginx in host mode.")
+	nginx.Flags().Bool("default-ingress", true, "Is this the default ingressClass for the cluster?")
```
Shikachuu, line 31: If we set this as default isn't this a backward incompatible change?

I think it would be better with default `false`.

Shikachuu (changes requested):

cmd/apps/nginx_app.go
Jasstkn: but what's the point for it being false? i thought that idea is to preconfigure application to be up&fully functional for development/testing purposes.

Shikachuu: To be clear, your approach makes more sense, if we have no promise for backward compatibility (which afaik we doesn't) you are right, it is better this way then.

Shikachuu (approved):

alexellis:
Thanks both

Issue #941, opened 2023-06-22
Add aws-cli


winacid:
would be nice to have aws-cli in arkade

alexellis:
It'd be nice but it's not a static binary so it's hard to install.

The only option I can think of is for `arkade system install`, since it has its own bespoke installer that has to run:

https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html

Does someone want to volunteer for this or to become a sponsor and I'll do it for you?

VariableExp0rt:
@alexellis I'd be happy to give this a go, given the docs mentioning running additional commands to run the installer of choice, should that be in a `PostInstallationMsg()` of some variety?

alexellis:
Hi @VariableExp0rt you'd need to look at one of the examples of a system install command and copy that example.

Alex

VariableExp0rt:
Hey @alexellis, the installation for this one was a bit complex due to the bespoke installer AWS provide for certain platforms. Happy to continue cleaning up that PR with any changes 👍

VariableExp0rt:
Hey @alexellis, I believe the PR for this has been approved, but is awaiting approval for a couple of workflow runs. Let me know if everything is sufficient there 👍

PR #993, opened 2023-10-24
Add AWS CLI to `system install` tools 

VariableExp0rt:
## Description

This PR attempts to add support for downloading and installing the AWS CLI for supported platforms. This turned out to be a bit more difficult with the special requirements for each platform supported (Linux, MacOS, Windows).

## Motivation and Context

Fixes a currently open issue - https://github.com/alexellis/arkade/issues/941

## How Has This Been Tested?

Changes tested with:

```
gofmt -w -s ./pkg
gofmt -w -s ./cmd

go build

make test
make e2e
```

```
```

## Types of changes

## Documentation

- [x] I have updated the list of apps in README.md if (required) with `./arkade install --help`

## Checklist:

(I've added a boolean flag `--run-installer` which defaults to false)


alexellis:
@Jasstkn would you be able to take a look at the approach?

Jasstkn:
> @Jasstkn would you be able to take a look at the approach?

sure, will take a look. In the meantime, @VariableExp0rt can I ask you to add a confirmation how you tested the installation via arkade?

VariableExp0rt:
@Jasstkn I tested via `arkade system install aws-cli` (I haven't tested on the other platforms, as I'm primarily operating on a Mac and don't have access to a Windows box)

VariableExp0rt:
Hey @Jasstkn, let me know if there is anything else you need from me :)

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,317 @@
...
+		filename := filepath.Base(outPath)
+		if _, err = get.CopyFileP(
+			outPath,
+			fmt.Sprintf("%s/%s", workDir, filename), readWriteExecuteEveryone,
+		); err != nil {
+			return err
+		}
+
+		extension := filepath.Ext(outPath)
+		if extension == ".zip" {
```
Shikachuu, line 154: The `get` package already have a method for this on the `Tool` struct: `IsArchive` I would rather use that.

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+
+		arch, osVer := env.GetClientArch()
+
+		if cmd.Flags().Changed("arch") {
+			arch, _ = cmd.Flags().GetString("arch")
+		}
+
+		if version == "" {
+			v, err := getAWSCLIVersion("aws", "aws-cli")
```
Shikachuu, line 84: Can't we replace this function + call to `FindGitHubRelease` from the `get` package?

That'll also cut the `go-git` from the dependency list.

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,317 @@
...
+		Example: `  arkade system install aws-cli
+  arkade system install aws-cli --version <version>`,
+		SilenceUsage: true,
+	}
+
+	command.Flags().StringP("version", "v", "", "The version or leave blank to determine the latest available version")
+	command.Flags().String("path", "/usr/local/bin", "Installation path, where a aws cli subfolder will be created")
+	command.Flags().Bool("progress", true, "Show download progress")
+	command.Flags().StringP("work-dir", "w", "", "Working directory that installer files should be copied to (current directory if not supplied)")
+	command.Flags().Bool("run-installer", false, "Whether or not arkade should run the downloaded installer")
```
Shikachuu, line 42: Shouldn't this be `true` by default?

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,317 @@
...
+
+func runBundledInstaller(osVer string, workDir string, filename string, installPath string) error {
+	fmt.Printf("Running bundled installer from download URL\n")
+
+	cmd := ""
+	args := make([]string, 0)
+
+	switch osVer {
+	case "Darwin":
+		cmd = "sudo"
```
Shikachuu, line 236: I would much prefer to asking for an explicit `sudo` and check the privileges on the `PreRunE` section, then implicitly calling it in the command.

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,317 @@
...
+		pkgDir := fmt.Sprintf(" -pkg %s/%s", workDir, filename)
+		args = append(args, "installer", pkgDir, "-target /")
+	case "Linux":
+		cmd = "sudo"
```
Shikachuu, line 240: Same as in line 236.

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,317 @@
...
+	command := &cobra.Command{
+		Use:   "aws-cli",
+		Short: "Install AWS CLI",
+		Long:  `Install AWS CLI for interacting with Amazon Web Services APIs.`,
+		Example: `  arkade system install aws-cli
+  arkade system install aws-cli --version <version>`,
+		SilenceUsage: true,
+	}
+
+	command.Flags().StringP("version", "v", "", "The version or leave blank to determine the latest available version")
```
Shikachuu, line 38: The default should be `githubLatest` like in the other system commands.
Check the [`firecracker`  command](https://github.com/alexellis/arkade/blob/master/cmd/system/firecracker.go#L56-L65) for example.

Shikachuu (changes requested):
Please see my comments above and do a rebase!

Overall really nice work, I really like the tool unit test for this too, it is a cool idea!

cmd/system/aws_cli.go
VariableExp0rt: I believe there was a reason for this, and it is that the AWS CLI doesn't publish GitHub Releases, but only supplies tags that refer to a given release. I'd be happy to replace this with something else to get rid of that though? Maybe just raw HTTP Get to retrieve the tags?

cmd/system/aws_cli.go
VariableExp0rt: I've addressed this by adding a prompt that sets a `runElevated` variable that is checked for `true` before running the bundled installer, otherwise falling back to printing the instructions, is this sufficient or does it need further re-thinking?

cmd/system/aws_cli.go
VariableExp0rt: I've removed the dependency on `go-git`, but still the tags are the thing we'd need here I think?

Shikachuu: Hmm...you are right, the tags are waaay up-to-date then the releases page, releases should be the way to go.

Also I really like the API implementation!

cmd/system/aws_cli.go
Shikachuu: Nice idea, but I would stick to run the `sudo arakde system install aws-cli` much rather then making a prompt.

Seems like the other commands require this too [eg: firecracker](https://github.com/alexellis/arkade/blob/master/docs/arkade.tape#L58)

Shikachuu (changes requested):
Looks promising.

Please don't forget to update the [README.md](https://github.com/alexellis/arkade/blob/master/README.md?plain=1#L242-L254) and squash the commits in to single one, we like to keep the git history clean for single tools like this.

Thanks!

VariableExp0rt:
> Looks promising.
> 
> Please don't forget to update the [README.md](https://github.com/alexellis/arkade/blob/master/README.md?plain=1#L242-L254) and squash the commits in to single one, we like to keep the git history clean for single tools like this.
> 
> Thanks!

Have updated the README.md [here](https://github.com/alexellis/arkade/pull/993/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R244) 👍 .

VariableExp0rt:
@Shikachuu I'm having quite a bit of trouble squashing into a single commit right now, is there a correct way I should be doing that other than `git rebase -i HEAD~7` and `squash`-ing my commits? I think the previous rebase is making things awkward

VariableExp0rt:
@Shikachuu I think I've managed to fix it, sorry about that! Let me know if there are any more comments to address 👍

Shikachuu (approved):

VariableExp0rt:
For posterity:

```
./hack/test-tool.sh awscli
+ ./arkade get awscli --arch arm64 --os darwin --quiet
+ file /Users/liam.baker/.arkade/bin/awscli
/Users/liam.baker/.arkade/bin/awscli: xar archive compressed TOC: 4792, SHA-1 checksum
+ rm /Users/liam.baker/.arkade/bin/awscli
+ echo

+ ./arkade get awscli --arch x86_64 --os darwin --quiet
+ file /Users/liam.baker/.arkade/bin/awscli
/Users/liam.baker/.arkade/bin/awscli: xar archive compressed TOC: 4792, SHA-1 checksum
+ rm /Users/liam.baker/.arkade/bin/awscli
+ echo

+ ./arkade get awscli --arch x86_64 --os linux --quiet
+ file /Users/liam.baker/.arkade/bin/awscli
/Users/liam.baker/.arkade/bin/awscli: Zip archive data, at least v2.0 to extract, compression method=store
+ rm /Users/liam.baker/.arkade/bin/awscli
+ echo

+ ./arkade get awscli --arch aarch64 --os linux --quiet
+ file /Users/liam.baker/.arkade/bin/awscli
/Users/liam.baker/.arkade/bin/awscli: Zip archive data, at least v2.0 to extract, compression method=store
+ rm /Users/liam.baker/.arkade/bin/awscli
+ echo

+ ./arkade get awscli --arch x86_64 --os mingw --quiet
+ file /Users/liam.baker/.arkade/bin/awscli.exe
/Users/liam.baker/.arkade/bin/awscli.exe: Composite Document File V2 Document, Little Endian, Os: Windows, Version 6.2, MSI Installer, Code page: 1252, Title: Installation Database, Subject: AWS Command Line Interface v2, Author: Amazon Web Services, Keywords: Installer, Comments: This installer database contains the logic and data required to install AWS Command Line Interface v2., Template: x64;1033, Revision Number: {568A8D9E-6206-47B9-8AF8-EA26143A52B8}, Create Time/Date: Fri Oct 20 19:03:24 2023, Last Saved Time/Date: Fri Oct 20 19:03:24 2023, Number of Pages: 200, Number of Words: 2, Name of Creating Application: Windows Installer XML Toolset (3.11.1.2318), Security: 2
+ rm /Users/liam.baker/.arkade/bin/awscli.exe
+ echo
```

VariableExp0rt:
@Shikachuu is there anything else I should do here?

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,329 @@
...
+		return a.Compare(b)
+	}
+
+	slices.SortFunc(tags, comparer)
+	latest := tags[len(tags)-1]
+
+	return latest.String(), nil
+}
+
+func runBundledInstaller(osVer string, workDir string, filename string, installPath string) error {
```
alexellis, line 241: I don't understand why this option is being included?

What format is the AWS CLI in?

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+
+		return nil
+	}
+
+	return command
+}
+
+func getAWSCLIVersion(owner, repo string) (string, error) {
+	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/git/refs/tags", owner, repo)
```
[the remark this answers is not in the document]
VariableExp0rt: The reason for this is that tags exist that are for all releases, but there are not releases to accompany them. I might be wrong, but I couldn't see specifically anywhere that only looks at the refs/tags (which is all that AWS provide)

cmd/system/aws_cli.go
VariableExp0rt: The CLI comes in different formats for different platforms, for instance OSx is a .pkg, Linux is a .zip, and similar for Windows. There isn't just a binary.

cmd/system/aws_cli.go
```diff
@@ -0,0 +1,329 @@
...
+
+	if result.ExitCode != 0 {
+		return fmt.Errorf("error running bundled installer for platform, stderr: %s", result.Stderr)
+	}
+
+	return nil
+}
+
+func installationInstructions(osVer string, workDir string, filename string, installPath string) ([]byte, error) {
+	t := template.New("Installation Instructions")
```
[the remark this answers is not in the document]
VariableExp0rt: There is not, just zip, pkg, or the msi installer for Windows.

PR #1095, opened 2024-07-01
Use go template for powershell and golang

### LINKED ISSUE #1094: Use Go template for system install apps
As of now custom code is being written to support each system install app. We want to extend current go [template](https://github.com/alexellis/arkade/blob/master/pkg/get/tools.go) setup for system install apps too.

nitishkumar71:
## Description

We are extending go template usage for system install apps like powershell and go
## Motivation and Context

Closes #1094

## How Has This Been Tested?

```
```

## Types of changes

## Documentation

## Checklist:


pkg/get/tools.go
```diff
@@ -4255,5 +4266,67 @@ https://github.com/{{.Owner}}/{{.Repo}}/releases/download/{{.Version}}/{{.Repo}}
...
 			{{.Name}}_{{.VersionNumber}}_{{$osStr}}_{{$arch}}.{{$ext}}`,
 		})
+
+	tools = append(tools,
+		Tool{
+			Owner:       "go",
+			Repo:        "go",
+			Name:        "go",
+			SystemOnly:  true,
+			Version:     "1.22.4",
```
nitishkumar71, line 4276: Since go does not uses github or k8s version strategy, will it be correct to hard-code it?

pkg/get/url_checker_test.go
```diff
@@ -14,6 +14,7 @@ func Test_CheckTools(t *testing.T) {
 	tools := MakeTools()
 	toolsToSkip := []string{
 		"kumactl", // S3 bucket disallow HEAD requests
+		"flyctl",
```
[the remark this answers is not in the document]
alexellis: OK let me know when you have..

pkg/get/tools.go
alexellis: One option is to hard-code it, the other thing we could have a strategy for go and move the code from the system installer there. The version can already be determined.

@rgee0 what are your thoughts? Go is installed via `system install` with custom code, we are looking at using the template system instead.

pkg/get/url_checker_test.go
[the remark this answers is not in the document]
nitishkumar71: Removed

pkg/get/tools.go
nitishkumar71: I was thinking to execute go version detection code to be only executed when cli is actually installing the go. I am not sure, how to setup that in current code base.

rgee0: I'd suggest a similar approach to the exiting version strategies but using: https://go.dev/VERSION?m=text

nitishkumar71: > I'd suggest a similar approach to the exiting version strategies but using: https://go.dev/VERSION?m=text

This is implemented in system [install](https://github.com/alexellis/arkade/blob/master/cmd/system/go.go#L111-L142) but we are stuck around how to implement same in template approach

rgee0: Apologies I missed `The version can already be determined`, maybe im also missing some context, but you're using `GetDownloadURL` which will take you into `GetUrl` which is where the existing version strategies are implemented.  `tool` is available all the way through so a `goVersionStrategy` could be implemented in there in much the same way as k8s is.

rgee0:
Thats the approach I had in mind.  You can see how with any further version strategies we might benefit from looking at optimising the approach to them, but I think we're OK for the moment.

The only thought I had was whether there'd be greater safety in parsing the content returned by go.dev/VERSION through a regex to extract the version.  This would defend a situation where the content might change to add additional content that precedes the existing

nitishkumar71:
> Thats the approach I had in mind. You can see how with any further version strategies we might benefit from looking at optimising the approach to them, but I think we're OK for the moment.
> 
> The only thought I had was whether there'd be greater safety in parsing the content returned by go.dev/VERSION through a regex to extract the version. This would defend a situation where the content might change to add additional content that precedes the existing

Changes are done.

rgee0 (approved):

PR #1110, opened 2024-08-08
Move upgradeable check to using semVer/pre

rgee0:
## Description

The number of tag version edge cases requiring exception from being upgraded was increasing to the point where constantly adding to the exception list was no longer a viable approach.

This change alters how the upgradability of a tag is assessed by parsing the current and latest tags into SemVers and then assessing the `pre` field of the version to determine whether either of the tags are a special case.  In order to achieve this it is necessary to move `latest` out to be a case all of itself as this would not be nicely handled by NewVersion().

## Motivation and Context

## How Has This Been Tested?
Unit only.

```
```

## Types of changes

## Documentation

## Checklist:
- [ x] My code follows the code style of this project.
- [ x] I've read the [CONTRIBUTION](https://github.com/alexellis/arkade/blob/master/CONTRIBUTING.md) guide
- [x ] I have signed-off my commits with `git commit -s`

### DIFF
--- a/cmd/chart/upgrade.go
+++ b/cmd/chart/upgrade.go
@@ -207,6 +207,6 @@ func tagIsUpgradeable(currentTag, latestTag string) bool {
 	currentSemVer, _ := semver.NewVersion(currentTag)
 	latestSemVer, _ := semver.NewVersion(latestTag)
 
-	return latestTag != currentTag && len(latestSemVer.Prerelease()) == 0 && len(currentSemVer.Prerelease()) == 0
+	return latestSemVer.Compare(currentSemVer) == 1 && latestSemVer.Prerelease() == currentSemVer.Prerelease()
 
 }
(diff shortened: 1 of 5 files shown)


cmd/chart/upgrade.go
```diff
@@ -200,6 +200,13 @@ func updateImages(iName string, v bool) (bool, string, error) {
...
 
-	return latestTag != currentTag && !strings.Contains(strings.ToLower(latestTag), "-rc") && !strings.EqualFold(currentTag, "latest")
+	if strings.EqualFold(currentTag, "latest") {
+		return false
+	}
+
+	currentSemVer, _ := semver.NewVersion(currentTag)
+	latestSemVer, _ := semver.NewVersion(latestTag)
+
+	return latestTag != currentTag && len(latestSemVer.Prerelease()) == 0 && len(currentSemVer.Prerelease()) == 0
```
alexellis, line 210: What about the case where the value is `v0.15.1-rootless` and we want to go up to `v0.15.2-rootless`?

rgee0: That will not upgrade.  

To achieve that we would need to check that the current/latest prereleases match.

alexellis (approved):
LGTM

PR #1135, opened 2024-11-09
feat: add hold file for `chart upgrade`

### LINKED ISSUE #1132: Upgrading container image tags - hold feature
## Expected Behaviour

I want some container images to be updated, and others to be pinned.

i.e. `postgres:16` should stay at `16`, not go to `17` when it happens to be available.

## Current Behaviour

When a specific container image is actually wanted, even if newer ones are available, it gets upgraded anyway.

## Possible Solution

A new file adjacent to the values.yaml with some convention for its name could be read and used to know to ignore that YAML path / selector / image.

Given the following, you may want to have one upgrade, and the other hold, in which case you need the full path to the image tag.

```
psql-1:
  image: postgres:16

psql-backup:
  image: postgres:16
```

But more likely the image only appears once, and only needs one version, so perhaps the image itself could be specified.

```
psql:
  image: postgres:16
```

hold.txt - specifies path in YAML to field to hold one per line

```
psql.image
```

hold.txt - specifies the specific source tag to ignore one per line

```
postgres:16
```

hold.yaml - specifies the same format 1:1 as the values.yaml, and the presence of anything in it means it's ignored:

```yaml
psql:
  image: postgres:16
```

## Steps to Reproduce (for bugs)

1. Save the values.yaml sample above
2. Use the sample above and run `arkade chart upgrade`

See it upgraded from 16 to 17

## Context

We are maintaining postgresql for certain charts with images that do need updating regularly mixed with ones which do not like Postgresql, and this has become tricky to manage.

rgee0:
## Description
Adds a feature to the `chart upgrade` command such that images within the target file can be ignored during the upgrade if they are detailed in a hold file.  To be considered the file should follow the name of the yaml file passed by `-f` suffixed with `.hold`, e.g. upgrading `docker-compose.yaml` would look for a `docker-compose.yaml.hold`.

The hold file format is:

```
component1.image
db.image
...
nats.image
```

## Motivation and Context
Closes #1132

## How Has This Been Tested?
TBC

## Types of changes

## Documentation

## Checklist:

### DIFF
--- a/cmd/chart/upgrade.go
+++ b/cmd/chart/upgrade.go
@@ -1,6 +1,7 @@
 package chart
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"log"
@@ -23,6 +24,8 @@ type tagAttributes struct {
 	hasPatch  bool
 	original  string
 }
+
+const holdFileExt = "hold"
 
 func (c *tagAttributes) attributesMatch(n tagAttributes) bool {
 	return c.hasMajor == n.hasMajor &&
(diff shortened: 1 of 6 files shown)


cmd/chart/upgrade.go
```diff
@@ -96,9 +97,22 @@ Otherwise, it returns a non-zero exit code and the updated values.yaml file.`,
 		}
 
 		if verbose {
-			if len(filtered) > 0 {
-				log.Printf("Found %d images\n", len(filtered))
-			}
+			log.Printf("Found %d images\n", len(filtered))
```
rgee0, line 100: `filtered` will be non-zero as we return earlier if `len==0`

PR #1140, opened 2024-11-21
feat: improve `install` error when app doesnt exist

rgee0:
## Description

When trying to install an app that doesnt exist the error message would advise the user that the app doesnt exist and that they should run `install --help` to see what is available.

However, there may be situations where the user is using the wrong verb because they are trying to get a tool, rather than an app.  This change introduces a check when an app not found error is encountered to see whether a tool with the user provided name exists and advises them accordingly.

also adds associated test.

## Motivation and Context

discussed on the community call on 20/11

### Functional
Tried the helm example with isnt an app but will resolve to a tool.  Then tried `burt` which is neither app nor tool.
```
➜  arkade git:(addGetPrompt) make build                                                         
go build
➜  arkade git:(addGetPrompt) ./arkade install helm                                              
Error: no such app. helm is available as a tool, run "arkade get helm" to get it
➜  arkade git:(addGetPrompt) ./arkade install burt                                              
Error: no such app: burt, run "arkade install --help" for a list of apps
```

## Types of changes

## Documentation

## Checklist:

### DIFF
--- a/cmd/install_test.go
+++ b/cmd/install_test.go
@@ -1,0 +1,91 @@
+// Copyright (c) arkade author(s) 2022. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package cmd
+
+import (
+	"testing"
+
+	"github.com/alexellis/arkade/pkg/get"
+)
+
+type Tool = get.Tool
+
+func TestCheckForTool(t *testing.T) {
+	tests := []struct {
+		name       string
+		appName    string
+		tools      []Tool
+		want       string
+		expectFail bool
+	}{
+		{
+			name:    "Tool exists but expected is wrong",
+			appName: "kubectl",
+			tools: []Tool{
+				{Name: "kubectl"},
+				{Name: "helm"},
+				{Name: "k9s"},
+			},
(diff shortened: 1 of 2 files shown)


cmd/install_test.go
```diff
@@ -0,0 +1,91 @@
...
+			expected:   "no such app: kubectl, run \"arkade install --help\" for a list of apps",
+			expectFail: false,
+		},
+		{
+			name:    "Empty app name",
+			appName: "",
+			tools: []Tool{
+				{Name: "kubectl"},
+			},
+			expected:   "no such app: , run \"arkade install --help\" for a list of apps",
```
alexellis, line 77: If there is some code with expect, we should change it to want instead for the code style of want/got.

cmd/install_test.go
```diff
@@ -0,0 +1,91 @@
...
+			expectFail: false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			result := checkForTool(tc.appName, tc.tools)
```
alexellis, line 84: result => got (please)

Issue #1220, opened 2025-11-18
Make shebang of test-tool.sh more generic


jceb:
I'm using nixos and I'm running into issues with the hard-coded shebang in file `/hack/test-tool.sh`, because the path doesn't exist. I propose to use a more generic approach that depends on `/usr/bin/env`.

alexellis:
Thanks for the feedback.

Funnily enough that isn't something we've ever run into in ~ 5 years.. 

I'll take the PR to encourage your usage but if we run into issues we may have to revert it.

PR #1221, opened 2025-11-18
fix: shebang

jceb:
## Description

Change shebang to `/usr/bin/env` and reformat code with shellcheck.

## Motivation and Context

#1220 

## How Has This Been Tested?

I ran the scripts locally.

## Types of changes

## Documentation

## Checklist:

### DIFF
--- a/hack/hashgen.sh
+++ b/hack/hashgen.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-for f in bin/arkade*; do shasum -a 256 $f > $f.sha256; done
+for f in bin/arkade*; do shasum -a 256 "$f" > "$f".sha256; done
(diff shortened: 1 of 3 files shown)


alexellis:
> Shebang change: From #!/bin/sh to #!/usr/bin/env bash. This is good for portability, as /usr/bin/env locates bash in PATH. However, since the script uses bash-specific features (none apparent), and original was /bin/sh, but shasum is standard, it's fine.

What is the default behaviour of `/bin/sh` on modern MacOS? Should `/bin/sh` be retained? i.e. `#!/usr/bin/env sh` - I would think all systems have `sh` or an alias to it, but some may lack bash itself?

jceb:
`/bin/sh` works on NixOS. I'm not sure about macOS, however it feels like a shell that'll definitely be place. I just checked again the scripts and couldn't see any bashisms, so I changed the shebang.

PR #1258, opened 2026-06-16
Fix symlink path-traversal escape in UntarNested

welteki:
## Description

A malicious OCI image layer could craft a tar archive containing a symlink pointing outside the extraction root, then use a subsequent entry to write a file through it, escaping the install directory.

All three entry types (file, dir, symlink) now resolve the physical path via `EvalSymlinks` and verify it stays within the extraction root before writing. Symlinks whose resolved target escapes the root, including chained escapes, are rejected. `validRelPath` is tightened to reject bare `..` and paths ending in `/..`.

Symlink extraction is now also opt-in: `UntarNested` takes an `allowSymlinks` parameter and `arkade oci install` exposes it via a new `--symlink` flag (default `false`), so symlinks in an OCI image are rejected unless the user explicitly enables them. System installers that rely on symlinks (e.g. Node.js) keep extracting them.

## Motivation and Context

The `UntarNested` function is used when extracting OCI image layers and nested tarballs. Without this fix a crafted archive could escape the install directory.

## How Has This Been Tested?

Unit tests cover all escape vectors (absolute symlink write-through, relative symlink write-through, chained symlink escape, planted escaping symlinks) and valid cases (internal symlinks, write-through internal symlink dirs, parent dir auto-creation), plus rejection of symlinks when `--symlink` is disabled.

```
go test ./pkg/archive/... -v
ok  github.com/alexellis/arkade/pkg/archive  0.007s
```

Several commands that exercise the `UntarNested` path were run to verify no regressions:

- `arkade system install`
- `arkade oci install`
- `arkade get`

## Types of changes

## Documentation

## Checklist:

### DIFF
--- a/cmd/oci/install.go
+++ b/cmd/oci/install.go
@@ -56,6 +56,7 @@ OCI image.`,
 
 	command.Flags().BoolP("gzipped", "g", false, "Is this a gzipped tarball?")
 	command.Flags().Bool("quiet", false, "Suppress progress output")
+	command.Flags().Bool("symlink", false, "Write symlinks when unpacking OCI image, only use with trusted sources")
 
 	// Hide the deprecated --path flag
 	command.Flags().MarkHidden("path")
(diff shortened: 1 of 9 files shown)


alexellis:
Thanks to "tonghuaroot" for flagging this to us using the official disclosure process.

We rate it as a very low risk item, unlikely to happen - unless of course you decide to run `arkade oci install` with a malware image from a third party.

PR #1261, opened 2026-07-02
Add restic backup tool with bzip2 decompression support

alexellis:
## Description

Add restic backup tool with bzip2 decompression support

Restic ships as .bz2 compressed static binaries for Linux/macOS and .zip for Windows. The decompress() function only handled .tar.gz, .zip, .tgz, and bare .gz, so bzip2 support was needed via compress/bzip2 from the stdlib (Go 1.25+).

Add unbzip() to download.go mirroring ungzip(), extend isArchiveStr() in get.go, and wire up the .bz2 branch in decompress().

Added by Qwen 3.6 27B - it ran through the testing instructions and updated the README.


alexellis:
PR is now stale/merged by other means.

PR #1273, opened 2026-08-19
Add vzzn to arkade get

alexellis:
## What

Adds `vzzn`, a lean vision/OCR client for OpenFaaS's internal toilgate
gateway, to `arkade get`. Also bumps dependencies (`go get -u`) and
regenerates the vendor tree in the same change.

## Unit tests

`Test_DownloadVzzn` pins v0.0.1 and covers all OS/arch combos, including
the arch aliases and an unsupported case.

## Manual verification

`hack/test-tool.sh vzzn` output:

```
darwin/arm64:    Mach-O 64-bit arm64 executable
darwin/x86_64:   Mach-O 64-bit x86_64 executable
linux/x86_64:    ELF 64-bit LSB executable, x86-64, statically linked
linux/aarch64:   ELF 64-bit LSB executable, ARM aarch64, statically linked
mingw/x86_64:    PE32+ executable (console) x86-64, for MS Windows
```

## E2E

`make e2e` (go test ./pkg/get --tags e2e) passes:

```
ok  github.com/alexellis/arkade/pkg/get  2.901s
```

## Keeping workflows pinned

Any action pinned to `@master` cannot be bumped by `arkade gha bump`,
which only raises major version tags. The fix is to move it onto a
versioned tag once, for example:

```
- uses: actions/checkout@master
+ uses: actions/checkout@v1
```

after which `arkade gha bump` can keep it current from then on:

```
arkade gha bump
```

`arkade gha bump` walks every workflow under `.github/workflows/` (or a
single file with `-f`), reads the versioned tags, and writes back the
latest major version. Use `--write=false` to dry-run. As a result,
`e2e-url-checker.yml` now pins `actions/checkout@v7` and
`actions/setup-go@v7` to match `build.yml` and `publish.yml`.


alexellis:
Follow-up: arch aliases handled.

The URL template now matches both the canonical host forms (from
`uname -m`: x86_64, aarch64, armv7l) and the aliases arkade's
`ValidateArch` accepts (amd64, arm64), so manually-specified flags such
as `--arch arm64 --os linux` or `--arch amd64` resolve to the correct
asset instead of falling back to the Linux x86_64 binary. Unsupported
combinations now fail closed (a missing `vzzn-unsupported` asset) rather
than installing a binary for another platform.

Re-ran `hack/test-tool.sh vzzn` — all five canonical combos still pass:

```
darwin/arm64:    Mach-O 64-bit arm64 executable
darwin/x86_64:   Mach-O 64-bit x86_64 executable
linux/x86_64:    ELF 64-bit LSB executable, statically linked
linux/aarch64:   ELF 64-bit LSB executable, ARM aarch64, statically linked
mingw/x86_64:    PE32+ executable (console) x86-64
```

Unit tests pin v0.0.1 and cover the aliases plus an unsupported case.

